Issue #4, opened 2017-11-06
Message timestamp check


lukasz-glen:
There is code in Golem that checks if message is not expired: checks if message has proper timestamp. Move this code to this repo and provide automatic check while parsing messages. Compare it to the requirements stated in 
https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/view#heading=h.xaztjc7pfto8

PR #7, opened 2017-11-10
Message timestamp check #4

jiivan:
fixes: #4 
Migrated time checking logic from golem.core.

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -1,4 +1,5 @@
 import cbor2
+import datetime
 import hashlib
 import logging
 import pytz
@@ -6,9 +7,31 @@ import struct
 import time
 from typing import Optional
 
+from . import exceptions
 from . import serializer
+from . import settings
 
 logger = logging.getLogger('golem.network.transport.message')
+
+
+def verify_time(msg):
+    """ Verify message timestamp. If message is to old or have timestamp from
+    distant future raise TimestampError.
+    """
+    now = datetime.datetime.utcnow()
+    try:
+        msgdt = datetime.datetime.utcfromtimestamp(msg.timestamp)
+    except (TypeError, OSError, OverflowError):
+        raise exceptions.TimestampError()
+    delta = now - msgdt
+    delta_future = msgdt - now
+    logger.debug('msgdt %s Δ %s Δfuture %s', msgdt, delta, delta_future)
+    if delta > settings.MSG_TTL:
+        # self.disconnect(BasicSafeSession.DCROldMessage)
+        raise exceptions.MessageTooOldError()
+    if delta_future > settings.FUTURE_TIME_TOLERANCE:
+        # self.disconnect(BasicSafeSession.DCRWrongTimestamp)
+        raise exceptions.MessageFromFutureError()
 
 
 # Message types that are allowed to be sent in the network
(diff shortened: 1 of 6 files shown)


tests/test_basic.py
```diff
@@ -1,14 +1,96 @@
...
+@freeze_time(testnow)
+class TimestampTestCase(unittest.TestCase):
+    """Time limits verification
+
+    Based on Concent_analiza_integracji_PL Limity czasu w komunikacji chapter
+    """
+
+    def setUp(self):
+        self.ecc = golem_messages.ECCx(None)
+        # mmtt - Maximum Message Transport Time, maksymalny dopuszczalny
```
etam: translate

tests/test_basic.py
```diff
@@ -1,14 +1,96 @@
...
         msg2 = golem_messages.load(payload, self.ecc.raw_privkey, self.ecc.raw_pubkey)
         self.assertEqual(msg, msg2)
+
+testnow = datetime.datetime.utcnow().replace(microsecond=0)
+
+@freeze_time(testnow)
+class TimestampTestCase(unittest.TestCase):
+    """Time limits verification
+
+    Based on Concent_analiza_integracji_PL Limity czasu w komunikacji chapter
```
etam: put quotes around chapter title

tests/test_basic.py
```diff
@@ -1,14 +1,96 @@
...
+        #        ten czas to komunikacja jest niedrożna).
+        self.mmtt = datetime.timedelta(minutes=0, seconds=30)
+        # mtd - Maximum Time Difference, maksymalne dopuszczalne odchylenie
+        #       czasu od czasu rzeczywistego.
+        self.mtd = datetime.timedelta(minutes=2, seconds=30)
+        # mat - Maximum Action Time, maksymalny czas na wykonanie prostej
+        #       operacji na maszynie.
+        self.mat = datetime.timedelta(minutes=2, seconds=15)
+
+    def test_timestamp_within_range(self):
```
etam: This contains 3 independent test cases, right? If so, they should be in separate methods.

tests/test_basic.py
```diff
@@ -1,14 +1,96 @@
...
+        msg = message.MessagePing()
+
+        # Message from the future
+        now = datetime.datetime.utcnow()
+        msg.timestamp = dt_to_ts(now + (self.mtd * 2) + one_second)
+        with self.assertRaises(exceptions.MessageFromFutureError):
+            message.verify_time(msg)
+
+    @mock.patch('golem_messages.message.verify_time')
+    def test_desserialization_with_time_verification(self, vft_mock):
```
etam: des-s-er

golem_messages/message.py
```diff
@@ -1,16 +1,37 @@
...
 from typing import Optional
 
+from . import exceptions
 from . import serializer
+from . import settings
 
 logger = logging.getLogger('golem.network.transport.message')
 
 
+def verify_time(msg):
```
etam: Maybe instead of passing whole message, just pass timestamp?
With this you could check it before constructing message object in deserialize.

etam: And you could test it without creating message objects.

Issue #2, opened 2017-11-03
CI


Elfoniok:
We need CI for this.

cameel:
What CI are you using in the rest of Golem? CircleCI?

maaktweluit:
We use:
- Buildbot ( all are migrated here in the future )
- CircleCI ( linux )
- Appveyor ( windows )
- Travix ( MacOS )
- Shippable ( lint on linux )

I recommend using CircleCI for now, its the most pleasant / stable of the list above.

PR #8, opened 2017-11-10
CircleCI

jiivan:
fixes: #2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,6 @@ jobs:
        - image: circleci/python:3.5
      steps:
        - checkout
-       - run:
-           command: |
-             ./setup.py test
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,19 @@
...
+       - image: circleci/python:3.5
+     steps:
+       - checkout
+       - restore_cache:
+           key: deps1-{{ .Branch }}-{{ checksum "setup.py" }}
+       - run:
+           command: |
+             python3 -m venv venv
+             . venv/bin/activate
+       - run: sudo apt-get install -y libgmp3-dev
```
maaktweluit: would do this as first command, to split between
- install global dependencies
- setup venv
- run

Issue #10, opened 2017-11-13
Field Reason in messages


lukasz-glen:
In many messages in documentation 
https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/view?pli=1#heading=h.dhlbf3h55z2u
there is a field Reason. Add Reason to the message MessageRejectReportComputedTask. Decide what is the best way to keep it: enum, stings, dictionary. Take into account existing structures in such messages as MessageCannotComputeTask, MessageTaskFailure, maybe others. Apply this decision to existing messages like MessageCannotComputeTask, MessageTaskFailure, MessageSubtaskResultsRejected according to the documentation. Reason options that do not yet exist need no new logic.

jiivan:
I think it should be simple string. Eventually a frozenset of allowed reason values may be used.
A simple string is readable and easy to understand and debug.

Using dict (or any other custom mapping such as enum) only adds unnecessary middle layer, and is consequently less readable. The msg size benefit is negligible in context of potential resources size and minimal amount of such messages.

etam:
@jiivan Using enum in python is quite readable.

If you have
```python
class MyEnum(Enum):
  A = 1
  B = 2
  C = 3
```

Then, when debugging, you'll see not just a numeric value, but an object representation like this:

```<MyEnum.A: 1>```

Also, if while deserializing input, there is an illegal value, enum constructor will throw an exception. For a set of strings you'll need to manually check the value. It's not a big problem (just a line or two), but still something to care about.

So I'm not convinced, that using an enum is less readable.

**Comment by jiivan** [comment]
@etam good point, but it still doesn't address magic numbers issue. What I mean is that for example in logs/wireshark etc we can see:

```python
>>> from golem_messages import message
>>> m = message.MessageDisconnect()  # numerical magic number
>>> m.serialize(lambda x: b'*'*message.Message.SIG_LEN)
b'\x00\x02\x00\x05^\x1c\xb6\xf2\xf3&\x00*****************************************************************\xd8\x1c\x81\xd8\x1c\x82freason '
```

I think this is easier to debug:
```python
>>> m = message.MessageDisconnect('TooManyPeers')  # Descriptive label
>>> m.serialize(lambda x: b'*'*message.Message.SIG_LEN)
[...]

PR #12, opened 2017-11-13
added Message{Ack,Reject}ReportComputedTask

etam:
fixes golemfactory/golem#1626

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -991,6 +991,57 @@ class MessageSubtaskPaymentRequest(Message):
         super(MessageSubtaskPaymentRequest, self).__init__(**kwargs)
 
 
+class MessageAckReportComputedTask(Message):
+    TYPE = TASK_MSG_BASE + 29
+
+    __slots__ = [
+        'subtask_id',
+    ] + Message.__slots__
+
+    def __init__(self, subtask_id=None, **kwargs):
+        self.subtask_id = subtask_id
+        super().__init__(**kwargs)
+
+
+class MessageRejectReportComputedTask(Message):
+    TYPE = TASK_MSG_BASE + 30
+
+    @enum.unique
+    class Reason(enum.Enum):
+        """
+        since python 3.6 it's possible to do this:
+
+        class StringEnum(str, enum.Enum):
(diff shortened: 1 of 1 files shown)


jiivan:
https://github.com/golemfactory/golem-messages/issues/10#issuecomment-344070094

golem_messages/message.py
```diff
@@ -1137,6 +1138,44 @@ def __init__(self, subtask_id=None, **kwargs):
...
+        self.subtask_id = subtask_id
+        super().__init__(**kwargs)
+
+
+@unique
+class MessageRejectReportComputedTaskReason(Enum):
+    TASK_TIME_LIMIT_EXCEEDED = 1
+    SUBTASK_TIME_LIMIT_EXCEEDED = 2
+    GOT_MESSAGE_CANNOT_COMPUTE_TASK = 3
+    GOT_MESSAGE_TASK_FAILURE = 4
```
jiivan: Since this enum is tightly coupled with `MessageRejectReportComputedTask` could you move it inside that class and name it `REASON`? Also according to discussion in #10 I will appreciate using more sophisticated labels than 1,2,3,4 :)

golem_messages/message.py
```diff
@@ -1330,6 +1369,8 @@ def init_messages():
             MessageSubtaskResultAccepted,
             MessageSubtaskResultRejected,
             MessageDeltaParts,
+            MessageAckReportComputedTask,
+            MessageRejectReportComputedTask,
```
jiivan: It's just a suggestion, but please give it a thought. Maybe it's worth creating commong abstract class with `subtask_id`?
```
AbstractReportComputedTask (`__slots__` = ("subtask_id",))
*-> MessageAckReportComputedTask (`__slots__` = ())
*-> MessageRejectReportComputedTask (`__slots__` = ("reason`,)

jiivan: Any tests for this?

golem_messages/message.py
etam: done

golem_messages/message.py
etam: done

Issue #6, opened 2017-11-09
MessageTaskToCompute is not fully defined


cameel:
The comments in the definition of `MessageTaskToCompute` refer to `ComputeTaskDef` but this class is only defined in [taskbase.py in golem repo](https://github.com/golemfactory/golem/blob/0.9.0/golem/task/taskbase.py#L152-L167). Its definition is needed to fully decode the payload of the message. I think it should be moved to this repository.

lukasz-glen:
Agree.
golem-messages must not have such dependencies.
To be done: check usages of ComputeTaskDef and if there loosely coupled with the rest of the Golem code then move it to golem-messages. If it is tightly coupled then we must figure out how to do it properly.

PR #17, opened 2017-11-16
Migrate ComputeTaskDef to golem-messages.

jiivan:
fixes: #6

As `ComputeTaskDef` was already only a mapping container I've also re factored it to implement mapping interface. It also allowed to catch implementation errors (invalid attributes).

### DIFF
--- a/golem_messages/datastructures.py
+++ b/golem_messages/datastructures.py
@@ -1,0 +1,20 @@
+import copy
+
+
+class FrozenDict(dict):
+    """FrozenDict allows only values provided in ITEMS attribute. It also acts
+       as DefaultDict with values from ITEMS.
+       """
+
+    ITEMS = {}
+
+    def __missing__(self, key):
+        return copy.deepcopy(self.ITEMS[key])
+
+    def __setitem__(self, key, value):
+        if key not in self.ITEMS:
+            raise KeyError("Invalid key: {}".format(key))
+        return super().__setitem__(key, value)
+
+    def __setattr__(self, key, value):
+        raise AttributeError("Read only. Use mapping interface")
--- a/golem_messages/datastructures.py
+++ b/golem_messages/datastructures.py
@@ -2,7 +2,7 @@ import copy
 
 
 class FrozenDict(dict):
-    """FrozenDict allows only values provided in ITEMS attribute. It also acts
+    """FrozenDict allows only keys provided in ITEMS attribute. It also acts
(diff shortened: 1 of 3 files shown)


golem_messages/message.py
```diff
@@ -34,6 +34,44 @@ def verify_time(timestamp):
         raise exceptions.MessageFromFutureError()
 
 
+class FrozenDict(dict):
```
etam: https://pypi.python.org/pypi/frozendict ?

jiivan: Same name different functionality :)

etam: How about [\_\_slots\_\_](https://stackoverflow.com/questions/472000/usage-of-slots) with default values?

Here's an example (from https://mail.python.org/pipermail/python-dev/2003-May/035607.html):

```python
>>> class Pane(object):
	__slots__ = { 'background': 'black', 'foreground': 'white',
		      'size': (80, 25) }
	def __init__(self):
		for k, v in self.__slots__.iteritems():
			setattr(self, k, v)

			
>>> p = Pane()
>>> p.background = 'blue'
>>> p.background, p.foreground
('blue', 'white')
```

jiivan: It's a nice hack of `__slots__` but it doesn't have the advantage of dictcless instance (still a dict for default values). And it's a hack... I prefer for dicts to be straightforward and look like dicts.

etam: So lets do it this way:

- Add some documentation to express what this FrozenDict is about.
- move it to a separate file.
- change `__missing__` to 

```python
from copy import deepcopy

def __missing__(self, key):
    return deepcopy(self.ITEMS[key])
```
This will allow to put dict or anything else as a default value.

etam: So lets do it this way:

- Add some documentation to express what this FrozenDict is about.
- move it to a separate file.
- change `__missing__` to 

```python
from copy import deepcopy

def __missing__(self, key):
    return deepcopy(self.ITEMS[key])
```
This will allow to put dict or anything else as a default value.

golem_messages/datastructures.py
```diff
@@ -0,0 +1,20 @@
+import copy
+
+
+class FrozenDict(dict):
+    """FrozenDict allows only values provided in ITEMS attribute. It also acts
```
etam: I'd change that to

> FrozenDict allows only keys provided in ITEMS. [...]

There are no restrictions on allowed values.

PR #18, opened 2017-11-17
Update gitignore

jiivan:


.gitignore
```diff
@@ -2,4 +2,8 @@
 *.egg-info
 __pycache__
 *.py[co]
-*.swp
+*.sw[ponm]
```
maaktweluit: This line looks developer specific, shouldn't it be in the users global gitignore? or are these files generated by the project?

No big deal if they stay, but when the file gets bigger it would keep it clean.

maaktweluit:
```
set backupdir=~/.vim-tmp,~/.tmp,~/tmp,/var/tmp,/tmp
set directory=~/.vim-tmp,~/.tmp,~/tmp,/var/tmp,/tmp
```
in vim-config will remove them from the working directory :)
https://coderwall.com/p/sdhfug/vim-swap-backup-and-undo-files

jiivan: Nice tip with backup dirs, still it's a custom configuration, and as I see it, adding *.swp and other ide related files to gitignore is a common practice.

PR #19, opened 2017-11-20
Reason field

jiivan:
fixes: #10

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -895,6 +895,14 @@ class MessageCannotComputeTask(Message):
         'subtask_id'
     ] + Message.__slots__
 
+    class REASON(enum.Enum):
+        WrongCTD = 'wrong_ctd'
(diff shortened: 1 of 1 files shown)


golem_messages/message.py
```diff
@@ -878,6 +895,14 @@ class MessageCannotComputeTask(Message):
         'subtask_id'
     ] + Message.__slots__
 
+    class REASON(enum.Enum):
+        WrongCTD = 'wrong_ctd'
+        WrongKey = 'wrong_key'
+        WrongAddress = 'wrong_address'
+        WrongEnvironment = 'wrong_environment'
```
maaktweluit: old message used to have env name in it too, is this data now lost?

jiivan: I don't think so. Environment is sent in `ComputTaskDef` by requestor, so he'll now what environment is problematic.

PR #22, opened 2017-11-21
Update config.yml

ggruszczynski:
thx to @maaktweluit

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,11 @@ jobs:
            command: |
              python3 -m venv venv
              . venv/bin/activate
-       - run: python3 setup.py test
+             pip install -r requirements.txt
+       - run:
+           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -12,7 +12,11 @@ jobs:
            command: |
              python3 -m venv venv
              . venv/bin/activate
-       - run: python3 setup.py test
+             pip install -r requirements.txt
```
jiivan: If adding `requirements.txt` is needed, then pls move them from `setup.py`.

ggruszczynski:
Can be reopened if needed.

Might be useful:
https://packaging.python.org/discussions/install-requires-vs-requirements/
https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/

PR #31, opened 2017-11-28
Messages should be initialized even before creating first message

Krigpl:
In case when we haven't sent any message yet, but are receiving a message from the network


etam:
It looks fine, but circleci test failed

tests/test_basic.py
```diff
@@ -27,6 +27,21 @@ def test_total_basic(self):
                                    self.ecc.raw_pubkey)
         self.assertEqual(msg, msg2)
 
+    """ Deserialization should work even if we haven't created any message first
+    """
```
jiivan: This docstring should probably be in `test_deserialization` method.

Issue #36, opened 2017-11-30
Enforce valid ComputeTaskDef


jiivan:
Enforce ComputeTaskDef type in `TaskToCompute.deserialize_slot()` and remove logic from https://github.com/golemfactory/golem/blob/develop/golem/task/tasksession.py#L716

depends on: #34

PR #38, opened 2017-12-01
Enforce valid ComputeTaskDef

jiivan:
fixes: #36

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -57,6 +57,18 @@ def _fake_sign(s):
     return b'\0' * Message.SIG_LEN
 
 
+def deserialize_task_to_compute(key, value):
+    if key == 'task_to_compute':
+        if not isinstance(value, (MessageTaskToCompute, type(None))):
+            raise TypeError(
+                "Invalid nested message type {} should be {}".format(
+                    type(value),
+                    MessageTaskToCompute
+                )
+            )
+    return value
+
+
 # Message types that are allowed to be sent in the network
 registered_message_types = {}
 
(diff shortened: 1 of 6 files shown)


tests/test_basic.py
```diff
@@ -169,3 +170,18 @@ def test_invalid_task_to_compute(self):
             s = msg.serialize()
             msg2 = message.Message.deserialize(s, decrypt_func=None)
             self.assertIs(msg2, None)
+
+
+class ComputeTaskDefTestCase(unittest.TestCase):
+    def test_type(self):
+        ctd = message.ComputeTaskDef()
+        ctd['src_code'] = ("Nie mamy w Polsce monarchy,"
```
maaktweluit: I cant review this part, should it be in polish?

golem_messages/message.py
```diff
@@ -697,6 +697,12 @@ def __init__(self, compute_task_def=None, **kwargs):
         self.compute_task_def = compute_task_def
         super().__init__(**kwargs)
 
+    def deserialize_slot(self, key, value):
+        value = super().deserialize_slot(key, value)
+        if key == 'compute_task_def':
+            value = ComputeTaskDef(value)
```
maaktweluit: what happens when there is an error? is this caught and logged on a higher level?

jiivan: Yes, in `.deserialize()` https://github.com/golemfactory/golem-messages/blob/master/golem_messages/message.py#L252

tests/test_basic.py
jiivan: :) It's just a poem by Tadeusz Boy-Żeleński. If you feel uncomfortable with this I can remove it. It does not affect the test.

maaktweluit: Cool! no need, maybe a comment above it that it is just test data / lorem ipum would be enough :)

mfranciszkiewicz:
Please create a release ASAP, currently it is not possible to compute tasks in develop.

Issue #14, opened 2017-11-13
MessageCannotComputeTask in MessageRejectReportComputedTask


lukasz-glen:
Implement including MessageCannotComputeTask in MessageRejectReportComputedTask: if Requestor have recieved message MessageCannotComputeTask for given substask, he must respond with MessageRejectReportComputedTask with embedded MessageCannotComputeTask  on MessageReportComputedTask.
Documentation:
https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/edit?pli=1#heading=h.dhlbf3h55z2u

depends on: #34

PR #41, opened 2017-12-01
MessageCannotComputeTask in MessageRejectReportComputedTask

jiivan:
fixes: #14

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -83,6 +83,12 @@ def deserialize_task_failure(key, value):
     return value
 
 
+def deserialize_cannot_compute_task(key, value):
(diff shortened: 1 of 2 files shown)


golem_messages/message.py
```diff
@@ -83,6 +83,12 @@ def deserialize_task_failure(key, value):
     return value
 
 
+def deserialize_cannot_compute_task(key, value):
+    if key == 'cannot_compute_task':
```
maaktweluit: will these string values only be used in this file?

If not it might be good to make it a constant or enum.

jiivan: They are defined in `__slots__` (key is an attribute name)

Issue #50, opened 2017-12-14
Signed message inside signed message gets malformed during serialization/deserialization


rwrzesien:
This code should explain the issue. Generally the signed message inside signed message gets malformed during serialization/deserialization.

``` python
from golem_messages.cryptography import ECCx
from golem_messages.cryptography import ecdsa_verify
from golem_messages.message import ComputeTaskDef
from golem_messages.message import ForceReportComputedTask
from golem_messages.message import TaskToCompute
from golem_messages.shortcuts import dump
from golem_messages.shortcuts import load


def generate_ecc_key_pair():
    ecc = ECCx(None)
    return ecc.raw_privkey, ecc.raw_pubkey


# Create 3 pairs of keys to emulate Concent flow
(CONCENT_PRIVATE_KEY,   CONCENT_PUBLIC_KEY)   = generate_ecc_key_pair()
(PROVIDER_PRIVATE_KEY,  PROVIDER_PUBLIC_KEY)  = generate_ecc_key_pair()
(REQUESTOR_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY) = generate_ecc_key_pair()


# Create and fill some data into ComputeTaskDef
compute_task_def = ComputeTaskDef()
compute_task_def['task_id'] = 1
compute_task_def['deadline'] = 1510912800

# Create TaskToCompute
task_to_compute = TaskToCompute(compute_task_def = compute_task_def)

# Dump TaskToCompute to make it signed
dumped_task_to_compute = dump(task_to_compute, REQUESTOR_PRIVATE_KEY, PROVIDER_PUBLIC_KEY)

# Load TaskToCompute back to its original format
task_to_compute = load(dumped_task_to_compute, PROVIDER_PRIVATE_KEY, REQUESTOR_PUBLIC_KEY)

# Task TaskToCompute is verified correctly
ecdsa_verify(REQUESTOR_PUBLIC_KEY, task_to_compute.sig, task_to_compute.get_short_hash())

# Create ForceReportComputedTask and attach TaskToCompute to it
force_report = ForceReportComputedTask()
force_report.task_to_compute = task_to_compute

# Dump ForceReportComputedTask to make it signed
dumped_force = dump(force_report, PROVIDER_PRIVATE_KEY, CONCENT_PUBLIC_KEY)

# Load TaskToCompute back to its original format
loaded_force = load(dumped_force, CONCENT_PRIVATE_KEY, PROVIDER_PUBLIC_KEY)

[...]

PR #51, opened 2017-12-14
Signed message inside a signed message

jiivan:
fixes: #50

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -96,7 +96,7 @@ registered_message_types = {}
 class Message():
     """ Communication message that is sent in all networks """
 
-    __slots__ = ['timestamp', 'encrypted', 'sig', '_payload', '_raw']
+    __slots__ = ['timestamp', 'encrypted', 'sig', '_raw']
 
     TS_SCALE = 10 ** 6
     HDR_LEN = 11
@@ -125,7 +125,6 @@ class Message():
         self.sig = sig
 
         # Encoded data
-        self._payload = payload  # child's payload only (may be encrypted)
         self._raw = raw  # whole message
 
     def __eq__(self, obj):
@@ -140,13 +139,23 @@ class Message():
         """Returns a raw copy of the message"""
         return self._raw[:]
 
-    def get_short_hash(self):
+    def get_short_hash(self, payload=None):
         """Return short message representation for signature
         :return bytes: sha1(TYPE, timestamp, encrypted, payload)
         """
+        if payload is None:
+            payload = serializer.dumps(self.slots())
         sha = hashlib.sha1()
-        sha.update(self.serialize_header())
-        sha.update(self._payload or b'')
+
+        # We can't use self.serialize_header() because it includes
+        # self.encrypted. And nested messages are decrypted, but they
+        # still need to have a valid signature.
+        # SEE: test_serializer.MessageTestCase.test_message_sig()
+        hash_header = serializer.dumps(
+            [self.TYPE, int(self.timestamp * self.TS_SCALE), ]
+        )
+        sha.update(hash_header)
+        sha.update(payload or b'')
         return sha.digest()
 
     def serialize(self, sign_func=None, encrypt_func=None):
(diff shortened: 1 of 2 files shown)


Krigpl:
I think this is fine but I feel like @mfranciszkiewicz should take a look at this as well.

tests/test_serializer.py
```diff
@@ -30,3 +32,67 @@ def test_message_list(self):
...
+
+        second_sig = force_report.task_to_compute.sig
+        second_hash = force_report.task_to_compute.get_short_hash()
+
+        self.assertEqual(first_sig, second_sig)
+        self.assertEqual(first_hash, second_hash)
+
+        # Now, attached TaskToCompute should still be verified using
+        # original key
+        cryptography.ecdsa_verify(
```
Krigpl: Is it supposed to throw if it's not signed properly? Can't see any asserts so it's not clear. Can you either assert that an exception is not thrown or add a comment?

jiivan: Not throwing an exception in unittest is default behaviour, you can't assert that. :)
https://github.com/golemfactory/golem-messages/blob/master/golem_messages/cryptography.py#L96

Not sure what comment can I add. `ecdsa_verify` is really short and rather straightforward.

Krigpl: What I meant is that without knowing (or looking into) the internals of `ecdsa_verify` you don't know that it throws on failure. For a person looking at the code it could be returning a boolean instead. So a comment saying that it will throw helps to understand that this isn't in fact no-op. Or doing something verbose like
```
thrown = False
try:
  foo()
except:
  thrown = True
assert not thrown
```

golem_messages/message.py
```diff
@@ -140,13 +139,23 @@ def raw(self):
         """Returns a raw copy of the message"""
         return self._raw[:]
 
-    def get_short_hash(self):
+    def get_short_hash(self, payload=None):
         """Return short message representation for signature
         :return bytes: sha1(TYPE, timestamp, encrypted, payload)
         """
+        if payload is None:
+            payload = serializer.dumps(self.slots())
```
mfranciszkiewicz: This is a step back and introduces a constant overhead. It means that when we sign message's hash, we serialize the already serialized slots. We also serialize slots of a message when we want to verify the signature, when the message was deserialized a moment ago.

golem_messages/message.py
```diff
@@ -140,13 +139,23 @@ def raw(self):
...
         sha = hashlib.sha1()
-        sha.update(self.serialize_header())
-        sha.update(self._payload or b'')
+
+        # We can't use self.serialize_header() because it includes
+        # self.encrypted. And nested messages are decrypted, but they
+        # still need to have a valid signature.
+        # SEE: test_serializer.MessageTestCase.test_message_sig()
+        hash_header = serializer.dumps(
```
mfranciszkiewicz: `serialize_header` can be a `classmethod`, taking `timestamp` and `encrypted` as arguments.

golem_messages/message.py
```diff
@@ -96,7 +96,7 @@ def deserialize_cannot_compute_task(key, value):
 class Message():
     """ Communication message that is sent in all networks """
 
-    __slots__ = ['timestamp', 'encrypted', 'sig', '_payload', '_raw']
+    __slots__ = ['timestamp', 'encrypted', 'sig', '_raw']
```
mfranciszkiewicz: `payload` is still passed to `__init__` (line 274)

jiivan: Good point. Created #54

golem_messages/message.py
jiivan: This would mean storing original `encrypted` value when serializing nested message with cbor. Sounds overcomplicated to me.

golem_messages/message.py
jiivan: It won't happen when signing (the `payment` arg is passed). It could be optimized in verification stage, but at this moment I think it's negligible. And removing `_payload` cache slot saves us from all cache related troubles.

mfranciszkiewicz: Golem was profiled and the overhead of re-serialization wasn't negligible. That's why this change was introduced.

jiivan: #55 Thanks for pointing this out.

Issue #55, opened 2017-12-20
Don't reserialize slots on immidiate message verification


jiivan:
Golem was profiled and the overhead of re-serialization wasn't negligible.

PR #56, opened 2017-12-20
Don't reserialize slots on immidiate message verification

jiivan:
fixes: #55
fixes: #54


tests/test_basic.py
```diff
@@ -42,6 +43,21 @@ def test_deserialization(self, verify_time):
         assert deserialized is not None
         assert deserialized.TYPE == message.Ping.TYPE
 
+    @mock.patch('golem_messages.serializer.dumps', wraps=serializer.dumps)
+    def test_slots_reselialization_optimization(self, dumps_mock):
+        """Don't reserialize message slots immidiately after deserialization"""
```
Krigpl: `s/immidiately/immediately/`

PR #59, opened 2017-12-29
Provide payment timestamp in SubtaskResultAccepted

Krigpl:
`reward` isn't used nor useful anyway.
Payment timestamp is needed for the provider to correctly identify all the payments requestor has paid for during `batchTransfer` (all payments with timestamp <= `closure_time`)

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -873,17 +873,15 @@ class SubtaskResultAccepted(Message):
 
     __slots__ = [
         'subtask_id',
-        'reward'
+        'payment_ts'
     ] + Message.__slots__
 
-    def __init__(self, subtask_id=0, reward=0, **kwargs):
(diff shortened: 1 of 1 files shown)


golem_messages/message.py
```diff
@@ -883,17 +883,15 @@ class SubtaskResultAccepted(Message):
 
     __slots__ = [
         'subtask_id',
-        'reward'
+        'payment_ts'
     ] + Message.__slots__
 
-    def __init__(self, subtask_id=0, reward=0, **kwargs):
+    def __init__(self, subtask_id: str="", payment_ts: int=0, **kwargs):
```
jiivan: Since the signature changes Is it possible to drop this `__init__` as in `ForceReportComputedTask`?

Krigpl: Discussed offline, decided to go with #61

PR #63, opened 2017-12-29
Handle empty keys in shortcuts

jiivan:


golem_messages/message.py
```diff
@@ -260,7 +260,11 @@ def deserialize(cls, msg, decrypt_func, check_time=True, verify_func=None):
             logger.info("Message error: invalid data: %r", exc)
             return
         if verify_func is not None:
-            verify_func(instance.get_short_hash(data), sig)
+            try:
+                verify_func(instance.get_short_hash(data), sig)
+            except Exception:
+                logger.warning('Failed to verify signature: %r', instance)
```
Krigpl: I dunno this is useful, since you rethrow the exception it's gonna be caught and logged somewhere else anyway

jiivan: At this stage `instance` is available. It can be useful for debugging. Will decrease severity.

PR #69, opened 2018-01-04
Initialize slots with None

jiivan:
* all message slots are initialized to `None` to avoid `AttributeErrors`
   in client code
 * slot values have preference over kwarg values (in other words kwarg
   values are treated as defaults, and values received over the wire are
   prioritized)

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -102,12 +102,13 @@ class Message():
         :param raw: original message bytes
         """
 
-        # Set attributes
-        for key in kwargs:
-            setattr(self, key, kwargs[key])
-
         # Child message slots
         self.load_slots(slots)
+
+        # Set attributes
+        for key in kwargs:
+            if getattr(self, key, None) is None:
+                setattr(self, key, kwargs[key])
 
         # Header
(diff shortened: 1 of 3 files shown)


golem_messages/message.py
```diff
@@ -351,8 +355,9 @@ class Hello(Message):
     ] + Message.__slots__
 
     def __init__(self, **kwargs):
-        self.golem_messages_version = golem_messages.__version__
         super().__init__(**kwargs)
+        if self.golem_messages_version is None:
```
maaktweluit: Does this mean no version received?
In that case would it be safer to set the version to 0.0.0?

jiivan: It can mean two things: 
 1. either no version received (no slot information or `None` in slot)
 1. a new message created

It could be hard to differentiate between these two cases. Assuming no malicious messages, empty value means older protocol id, and connection will be dropped. Assuming malicious message it doesn't really matter what we put there.

jiivan: After giving it a second thought I found a way to address this issue in a clean way

maaktweluit: Looks great! thank you for the reply and fix :)

Issue #70, opened 2018-01-05
InvalidSignature sometimes raised when decoding a message with a FrozenDict on Python 3.5.2


Jakub89:
There is a problem with using `datastructures.FrozenDict` with our `FileTransferToken`. It produce un-deterministic errors `golem_messages.exceptions.InvalidSignature`. One idea is that dict changes the order of those three values. Also that bug appear when tested on Python 3.5.2, but disappear with new 3.6.
Script we use for testing:
```
import datetime

from golem_messages.shortcuts       import dump, load
from golem_messages.message         import FileTransferToken, TokenFilesData


CONCENT_PUBLIC_KEY  = b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;'
CONCENT_PRIVATE_KEY = b'l\xcdh\x19\xeb$>\xbcG\xa1\xc7v\xe8\xd7o\x0c\xbf\x0e\x0fM\x89lw\x1e\xd7K\xd6Hnv$\xa2'

message_timestamp = 1546171200
token                                 = FileTransferToken(timestamp=message_timestamp)
token.token_expiration_deadline       = message_timestamp + 3600
token.storage_cluster_address         = 'http://storage.concent.golem.network/'
token.authorized_client_public_key    = b'\xf3\x97\x19\xcdX\xda\x86tiP\x1c&\xd39M\x9e\xa4\xddb\x89\xb5,]O\xd5cR\x84\xb85\xed\xc9\xa17e,\xb2s\xeb\n1\xcaN.l\xba\xc3\xb7\xc2\xba\xff\xabN\xde\xb3\x0b\xa6l\xbf6o\x81\xe0;'

token.files             = TokenFilesData()
token.files['path']     = 'blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'
token.files['checksum'] = '098f6bcd4621d373cade4e832627b4f6'
token.files['size']     = 1024
token.operation         = 'download'

serialize_token = dump(token, CONCENT_PRIVATE_KEY, CONCENT_PUBLIC_KEY)

[...]

PR #72, opened 2018-01-05
InvalidSignature sometimes raised

jiivan:
fixes: #70 

In python<3.6 dicts were serialized in different order on each interpreter startup. Changed serialization of dicts in cbor to always sort dict by keys, thus creating repeatable signatures.


cameel:
@Jakub89 Please test it on your machine.
@bartoszbetka It would also be a good idea to test it on the dev cluster.

PR #73, opened 2018-01-08
 Fix filling default values for list slots 

jiivan:
* Upgrade tests
 * Fix issues with golemfactory/golem#1858

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -448,14 +448,14 @@ class Tasks(Message):
 
     __slots__ = ['tasks'] + Message.__slots__
 
-    def __init__(self, tasks=None, **kwargs):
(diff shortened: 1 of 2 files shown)


golem_messages/message.py
```diff
@@ -454,7 +454,7 @@ def __init__(self, tasks=None, **kwargs):
         :param list tasks: list of tasks information (subset of
                            taskserver.get_tasks_headers())
         """
-        self.tasks = tasks or []
+        kwargs['tasks'] = tasks or []
```
mfranciszkiewicz: Is there any reason that this one is different from the others?

jiivan: No reason, just lost in development :). Fixed

PR #79, opened 2018-01-11
Helper script for deployment of sdist package.

jiivan:


jiivan:
@maaktweluit unfortunatelly distutils docs are not the most friendly, but in source of `distutils/commands/upload.py` you'll find:

```python
        # Sign if requested                                                     
        if self.sign:                                                           
            gpg_args = ["gpg", "--detach-sign", "-a", filename]                 
            if self.identity:                                                   
                gpg_args[2:2] = ["--local-user", self.identity]                 
            spawn(gpg_args,                                                     
                  dry_run=self.dry_run)
```

Thus you by default signature is made with `gpg` defaults. It can be overriden with `--identity`

Issue #81, opened 2018-01-12
Enable pylint


jiivan:
Enable `pylint` checks additionally to `pycodestyle`. Either as a `TestCase` or as `lintdiff` (as in golem core) if needed.

PR #82, opened 2018-01-12
Enable pylint test

jiivan:
fixes: #81


golem_messages/message/__init__.py
```diff
@@ -0,0 +1,146 @@
+# For backwards compatibility import all old messages by name
+# pylint: disable=unused-import
+# pylint: disable=cyclic-import
+from golem_messages.message.base import Message  # noqa
```
maaktweluit: maybe all those `# noqa` blocks can also be a `pylint: disable=` before the block

golem_messages/message/__init__.py
```diff
@@ -0,0 +1,146 @@
+# For backwards compatibility import all old messages by name
+# pylint: disable=unused-import
+# pylint: disable=cyclic-import
```
maaktweluit: should we enable these again after this block?

golem_messages/message/base.py
```diff
@@ -0,0 +1,382 @@
...
+import time
+
+import golem_messages
+
+from golem_messages import datastructures
+from golem_messages import exceptions
+from golem_messages import serializer
+from golem_messages import settings
+
+logger = logging.getLogger('golem.network.transport.message')
```
maaktweluit: no default `__name__` for all loggers?

golem_messages/message/base.py
```diff
@@ -0,0 +1,382 @@
...
+    delta = now - msgdt
+    delta_future = msgdt - now
+    logger.debug('msgdt %s Δ %s Δfuture %s', msgdt, delta, delta_future)
+    if delta > settings.MSG_TTL:
+        raise exceptions.MessageTooOldError()
+    if delta_future > settings.FUTURE_TIME_TOLERANCE:
+        raise exceptions.MessageFromFutureError()
+
+
+def _fake_sign(s):  # pylint: disable=unused-argument
```
maaktweluit: unused argument can be replaced with `_*`

golem_messages/message/base.py
```diff
@@ -0,0 +1,382 @@
...
+
+    HDR_LEN = 11
+    SIG_LEN = 65
+    PAYLOAD_IDX = HDR_LEN + SIG_LEN
+
+    TYPE = None
+    ENCRYPT = True
+    ENUM_SLOTS = {}
+
+    def __init__(self, timestamp=None, encrypted=False, sig=None,  # noqa TODO #88 pylint: disable=too-many-arguments
```
maaktweluit: nice todo!

golem_messages/message/__init__.py
jiivan: done

golem_messages/message/__init__.py
jiivan: if `pycodestyle` used block I would prefer block :)

golem_messages/message/base.py
jiivan: It was left like this for backwards compatibility (before moving `golem.network.transport.message` to `golem_messages.message` and then `golem_messages.message.base`)

golem_messages/message/base.py
jiivan: fixed

golem_messages/message/__init__.py
maaktweluit: damn you `pycodestyle`! ;)

Issue #77, opened 2018-01-11
Add Provider and Requestor to the message TaskToCompute


lukasz-glen:
From the Concent point of view the message TaskToCompute is an agreement between Requestor and Provider. It should contain Requetor's and Provider's golem id. There is a possiblility that field task_owner is Requestor's golem id or Provider's golem id is in headers, please check. Add logic in Golem that sets these fields (Requestor side) and checks if they are properly set (Provider side).
(edit) There should be Provider's Ethereum account in the message TaskToCompute also. It is needed because of Concent forced payments.

jiivan:
Research notes: `ComputeTaskDef.task_owner` is instance of `golem.network.p2p.node.Node` meaning Requestor node.

PR #83, opened 2018-01-12
Add Provider and Requestor to the message TaskToCompute

jiivan:
fixes: #77 

Backwards incompatible

### DIFF
--- a/golem_messages/message.py
+++ b/golem_messages/message.py
@@ -6,7 +6,6 @@ import hashlib
 import logging
 import struct
 import time
-from typing import Optional
 
 import golem_messages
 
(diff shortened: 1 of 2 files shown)


golem_messages/message.py
```diff
@@ -565,7 +566,24 @@ class WantToComputeTask(Message):
 class TaskToCompute(Message):
     TYPE = TASK_MSG_BASE + 2
 
-    __slots__ = ['compute_task_def'] + Message.__slots__
+    __slots__ = [
+        'requestor_id',
+        'provider_id',
+        'compute_task_def',
```
etam: What is this `compute_task_def`? Could you add a comment about it?

Issue #52, opened 2017-12-19
Catching DescryptionError in message.py


kbeker:
When we use `load()` method and we give wrong message we should get exception DescriptionError from golem_message. For this moment golem_messages doesn't catch this exception and goes on. After finish
`msg = message.Message.deserialize(data, decrypt, check_time)`

it continue its work and throw exception `AttributeError: 'NoneType' object has no attribute 'sig'` on next method inside `load()`:

`cryptography.ecdsa_verify(pubkey, msg.sig, msg.get_short_hash())`

but it should throw exception in `message.py`

```
except Exception as exc:
            logger.info("Message error: invalid data: %r", exc)
            logger.debug("Failing message hdr: %r data: %r", header, data)
            return
```

Above is place where this exception should be thrown.

jiivan:
I am also proponent of raising some kind of exception in this situation. Unfortunately it's more labor intensive than it may seem at first sight. All golem core codebase expects `None` after failed _deserialization_ and also lot of tests (both in golem core and golem-messages).

Still I second to that idea and hope to implement it in a near future.

PR #98, opened 2018-01-23
Raise instead of returning None

jiivan:
fixes: #52

### DIFF
--- a/golem_messages/cryptography.py
+++ b/golem_messages/cryptography.py
@@ -64,12 +64,17 @@ if 'pyelliptic' not in dir() \
     sys.exit(1)
 
 
+def verify_pubkey(key):
+    if len(key) != 64:
+        raise exceptions.InvalidKeys('Invalid pubkey length')
+
+
 def privtopub(raw_privkey):
     raw_pubkey = bitcoin.encode_pubkey(
         bitcoin.privtopub(raw_privkey),
         'bin_electrum'
     )
-    assert len(raw_pubkey) == 64
+    verify_pubkey(raw_pubkey)
     return raw_pubkey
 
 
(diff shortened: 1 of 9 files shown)


golem_messages/cryptography.py
```diff
@@ -317,5 +325,6 @@ def sign(self, data):
         return signature
 
     def verify(self, signature, message):
-        assert len(signature) == 65
+        if len(signature) != 65:
+            raise exceptions.INvalidSignature('Invalid length')
```
kbeker: Shouldn't be ` raise exceptions.InvalidSignature('Invalid length')'`  instead of `raise exceptions.INvalidSignature('Invalid length')`?

golem_messages/message/base.py
```diff
@@ -206,6 +212,8 @@ def deserialize_header(cls, data):
                 )
             )
 
+        from golem_messages.message import registered_message_types
+
```
kbeker: Wouldn't be better to have all imports in the top of file?

golem_messages/cryptography.py
jiivan: Already fixed in 9c30537 :)

golem_messages/message/base.py
jiivan: Cyclic import (`golem_messages.message.__init__.py`). You may want to have a look at #103

golem_messages/message/base.py
```diff
@@ -56,7 +56,7 @@ class Message():
 
     __slots__ = ['timestamp', 'encrypted', 'sig', '_raw']
 
-    HDR_LEN = 11
+    HDR_LEN = struct.calcsize('!HQ?')
```
maaktweluit: Not sure how often this is used, but maybe the string can be a const like variable too?

Issue #101, opened 2018-01-24
Remove serialization of custom objects


jiivan:
Remove all logic related with `golem_messages.serializer.encode_object()`

PR #106, opened 2018-01-24
Remove serialization of custom objects

jiivan:
fixes: #101


jiivan:
@Krigpl will be resolved after creating PR `Upgrade golem-messages to 1.7` in golem core.

See for example https://github.com/golemfactory/golem/pull/1974/

Krigpl:
> @Krigpl will be resolved after creating PR Upgrade golem-messages to 1.7 in golem core.

This is a weird order, usually you'd first remove usages and only when the offending code is no longer used anywhere you'd delete it.
What if it turns out that refactoring this dependency is more annoying than expected and will be delayed in time? And in the meantime some other feature gets added to this module and we'd like to update the version in golem core? We wouldn't be able to do that.

jiivan:
@Krigpl looks like there is no issue here, as nodelist sent in `Peers` is a list of dicts: https://github.com/golemfactory/golem/blob/develop/golem/network/p2p/p2pservice.py#L592

Krigpl:
Except that this value is a Python struct https://github.com/golemfactory/golem/blob/develop/golem/network/p2p/p2pservice.py#L588

jiivan:
@Krigpl good catch. Fixed in https://github.com/golemfactory/golem/pull/2013

golem_messages/serializer.py
```diff
@@ -17,7 +12,7 @@
 logger = logging.getLogger('golem.core.simpleserializer')
 
 
-OBJECT_TAG = 239
+OBJECT_TAG = 239  # unused, old protocol
```
Krigpl: Reason for keeping it?

Issue #88, opened 2018-01-17
Pass MessageHeader to Message constructor


jiivan:
Passing `datastructures.MessageHeader` to constructor would fix `pylint too-many-arguments`

Also storing header in `header` slot and removing `timestamp`, `encrypted` and `sig` slots would help organizing message information.

depends on: #98

PR #107, opened 2018-01-25
Pass MessageHeader to Message constructor

jiivan:
fixes: #88


jiivan:
@maaktweluit only in messages (and tests but this is haxing).

Issue #89, opened 2018-01-17
Refactor cryptography.py


jiivan:
Fix all warnings from pylint on `cryptography.py` and unignore it in `.pylintrc`

depends on: #98

PR #108, opened 2018-01-25
Delint cryptography.py

jiivan:
fixes: #89

### DIFF
--- a/golem_messages/cryptography.py
+++ b/golem_messages/cryptography.py
@@ -320,12 +320,12 @@ class ECCx(pyelliptic.ECC):
     encrypt = ecies_encrypt
     decrypt = ecies_decrypt
 
-    def sign(self, data):  # pylint: disable=arguments-differ
-        signature = ecdsa_sign(self.raw_privkey, data)
+    def sign(self, inputb):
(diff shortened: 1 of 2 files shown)


golem_messages/cryptography.py
```diff
@@ -319,12 +320,12 @@ def ecies_decrypt(self, data, shared_mac_data=b''):
     encrypt = ecies_encrypt
     decrypt = ecies_decrypt
 
-    def sign(self, data):
+    def sign(self, data):  # pylint: disable=arguments-differ
```
etam: Is this needed? The number of arguments is the same. It's just that arguments have different names.

golem_messages/cryptography.py
```diff
@@ -319,12 +320,12 @@ def ecies_decrypt(self, data, shared_mac_data=b''):
...
         signature = ecdsa_sign(self.raw_privkey, data)
         assert len(signature) == 65
         return signature
 
-    def verify(self, signature, message):
+    def verify(self, signature, message):  # pylint: disable=arguments-differ
```
etam: Like above.

Issue #95, opened 2018-01-19
Include golem-messages version in Message header


mfranciszkiewicz:
Version inclusion would allow to discard incompatible messages earlier and skip the deserialization / signature verification process.

jiivan:
depends on: #88

PR #109, opened 2018-01-26
Include golem-messages version outside of payload

jiivan:
fixes: #95


golem_messages/message/base.py
```diff
@@ -353,13 +397,55 @@ class Hello(Message):
         'challenge',
         'difficulty',
         'metadata',
+        '_version',
```
mfranciszkiewicz: Why not pass it along? As far as I can tell, slots starting with `_` are not initialized.

PR #113, opened 2018-01-29
Verify that task_id|subtask_id are at max 128 characters long

### LINKED ISSUE #99: Verify that task_id|subtask_id are at max 128 characters long
related to: https://github.com/golemfactory/concent/issues/55#issuecomment-359733851

jiivan:
Fixes #99

### DIFF
--- a/golem_messages/validators.py
+++ b/golem_messages/validators.py
@@ -1,0 +1,12 @@
+from golem_messages import exceptions
+
+
+def validate_varchar(field_name, value, max_length):
+    if not (isinstance(value, str) and len(value) <= max_length):
+        raise exceptions.FieldError(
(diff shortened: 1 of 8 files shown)


golem_messages/validators.py
```diff
@@ -0,0 +1,12 @@
+from golem_messages import exceptions
+
+
+def validate_varchar(field_name, value, max_length):
+    if not (isinstance(value, str) and len(value) <= max_length):
```
shadeofblue: would it make sense to also allow a `max_length` of `None` and then only check for the instance type?

jiivan: I'd add it if it'd needed.

PR #115, opened 2018-01-30
modify `SubtaskResultsRejected` message

### LINKED ISSUE #87: Add missing fields to SubtaskResultsRejected.
According to the [documentation](https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/edit#heading=h.dhlbf3h55z2u) message  SubtaskResultsRejected is missing two fields: inner message and reason. This issue has parts:

1. Add inner message. Small difficulty is that inner message has two possible types.
2. Add field reason as a dictionary. Remark: reasons i, vi, vii are set by Requestor, others are set by Concent.
3. Implement logic for reason i. It is simply response generated by Requestor's verification.
4. Implement logic for reason vii. Requestor needs to recognize if Concent is enabled. It is immediate response to Provider if download failed or hash is invalid.

Logic for reason vi is response to Concent request and should be implemented in other issue.

shadeofblue:
+ tests

fixes #87


golem_messages/message/tasks.py
```diff
@@ -162,10 +166,62 @@ class SubtaskResultsAccepted(base.Message):
...
+        #
+        from .concents import ForceGetTaskResultFailed
+        # pylint: enable=cyclic-import
+
+        value = super().deserialize_slot(key, value)
+        value = base.deserialize_verify(
+            key, value,
+            verify_key='report_computed_task',
+            verify_class=ReportComputedTask,
+        )
```
jiivan: `deserialize_report_computed_task` at the bottom

golem_messages/message/tasks.py
```diff
@@ -162,10 +166,62 @@ class SubtaskResultsAccepted(base.Message):
...
+        value = base.deserialize_verify(
+            key, value,
+            verify_key='force_get_task_result_failed',
+            verify_class=ForceGetTaskResultFailed,
+        )
```
jiivan: I would put this deserializer in `message.concents` as `deserialize_force_get_task_result_failed`

tests/test_messages.py
```diff
@@ -382,6 +382,52 @@ def test_task_to_compute_validate_compute_task_def(self):
             ))
 
 
+class TasksTest(unittest.TestCase):
```
jiivan: Maybe put it in `test_tasks` like in https://github.com/golemfactory/golem-messages/pull/113/ ?

golem_messages/message/tasks.py
shadeofblue: @jiivan hmm... generally, I find those helpers of dubious value... you're essentially just binding the parameter name with the class name but otherwise, not adding much to the code readability and/or DRY... but sure, I'll add that helper...

tests/test_messages.py
shadeofblue: @jiivan hmm... so maybe #113 should be merged first? ... but anyway, maybe we should add a subpackage `message` in `tests` and put those tests in there?

jiivan: good idea 👍

golem_messages/message/tasks.py
```diff
@@ -353,3 +335,24 @@ class SubtaskPaymentRequest(base.Message):
     verify_key='report_computed_task',
     verify_class=ReportComputedTask,
 )
+
+def deserialize_force_get_task_result_failed(key, value):
```
jiivan: Why not move this to `message.concents` as I mentioned? :)

shadeofblue: ah, oops...

shadeofblue: @jiivan moved

PR #126, opened 2018-02-05
Messages 111 force subtask results

### LINKED ISSUE #111: Create the message ForceSubtaskResults
Create the message ForceSubtaskResults with respect to the [documentation](https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/edit#heading=h.dhlbf3h55z2u).

shadeofblue:
Adds the `ForceSubtaskResults` message plus fills gaps in the documentation of the previously-added messages.

closes #111


tests/message/test_concents.py
```diff
@@ -0,0 +1,171 @@
...
+            slots__task_to_compute=ttc
+        )
+        expected = [
+            ['origin',
+             concents.SubtaskResultsSettled.Origin.ResultsAcceptedTimeout
+             .value],
+            ['task_to_compute', ttc]
+        ]
+
+        self.assertEqual(expected, msg.slots())
```
jiivan: You could use `self.assertCountEqual()` and don't care about the ordering of slots.

shadeofblue: @jiivan hmm, doesn't the equality check verify also the contents?

jiivan: @shadeofblue https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

Issue #132, opened 2018-02-08
Missing ForceSubtaskResults and ForceSubtaskResultsResponse in message/__init__.py


rwrzesien:
ForceSubtaskResults and ForceSubtaskResultsResponse are not being added to `registered_message_types`

jiivan:
It shouldn't. Old messages are imported only for backwards compatibility. New code should reference messages directly: `message.concents.ForceSubtaskResults`

jiivan:
Reopenning after talk with @shadeofblue . The problem is not with import, but with `registered_message_types`

rwrzesien:
@jiivan @shadeofblue Added info about missing ForceSubtaskResultsResponse too.

PR #135, opened 2018-02-08
Add concents.ForceSubtaskResults to __init__.py

Jakub89:
Resolves https://github.com/golemfactory/golem-messages/issues/132


shadeofblue:
@Jakub89 ah, I was thinking of also adding tests checking for existence of each those messages in the registered messages list so that we get used to doing that with each message...

Issue #116, opened 2018-02-01
Message ForceReportComputedTask cannot be resend


lukasz-glen:
A change has been done in the documentation. It was stated that messages are resend by Concent but it is technically incorrect to resend a message. We must provide additional wrapping messages. Changes:

1. Message ForceGetTaskResult can contain only ReportComputedTask, it cannot contain ForceReportComputedTask anymore. Please update golem's logic accordingly. 
2. The same as above for the message MessageSubtaskResultsRejected. It can contain only ReportComputedTask from now on. Please update golem's logic accordingly. 
3. Message ForceReportComputedTask became a wrapper for ReportComputedTask and it is used only for transport purposes. See the documentation for changes in the message structure. Requestor does not have to keep ForceReportComputedTask or ReportComputedTask as a Provider's report, a link to results and a proof. Requestor keeps only ReportComputedTask. Please update golem's logic accordingly. 
4. Add the message MessageForceReportComputedTaskResponse. All non-generic responses from Concent are MessageForceReportComputedTaskResponse for the forcing report use case from now on. See the documentation. Please update golem's logic accordingly. 
5. Please revisit the force report use case. The changes pointed above should be sufficient but make sure that the use case is implemented according to the documentation.
6. Please revisit the force results download use case. The changes in ForceGetTaskResult pointed above should be sufficient but make sure that the use case is implemented according to the documentation.

PR #144, opened 2018-02-14
Message ForceReportComputedTask cannot be resent

jiivan:
fixes: #116

### DIFF
--- a/golem_messages/message/concents.py
+++ b/golem_messages/message/concents.py
@@ -265,13 +265,11 @@ class ForceGetTaskResult(base.Message):
 
     __slots__ = [
         'report_computed_task',
-        'force_report_computed_task',
     ] + base.Message.__slots__
 
     def deserialize_slot(self, key, value):
         value = super().deserialize_slot(key, value)
         value = tasks.deserialize_report_computed_task(key, value)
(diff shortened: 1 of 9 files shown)


golem_messages/message/concents.py
```diff
@@ -457,6 +459,43 @@ class REASON(enum.Enum):
...
+
+    __slots__ = base.AbstractReasonMessage.__slots__ + [
+        # AckReportComputedTask sent from Requestor to Concent
+        # OR generated and signed by Concent
+        'ack_report_computed_task',
+        # RejectReportComputedTask sent from Requestor to Concent
+        'reject_report_computed_task',
+    ]
+
+    @base.verify_slot_list(
```
shadeofblue: @jiivan `_list` ?

golem_messages/message/concents.py
```diff
@@ -457,6 +459,43 @@ class REASON(enum.Enum):
...
+        'ack_report_computed_task',
+        AckReportComputedTask,
+    )
+    @base.verify_slot_list(
```
shadeofblue: @jiivan an here -> are you sure it's `_list`?

shadeofblue:
@jiivan maybe some tests to ensure that the verification works as expected?

Issue #150, opened 2018-02-16
Eth address in TaskToCompute


lukasz-glen:
1. Add two fields to the message TaskToCompute: requestor's and provider's ethereum public key.
2. Add two helper functions to the message TaskToCompute that return calculated ethrereum addresses of requestor and provider accordingly.
3. Modify golem logic, two new fields should be equal to requestor_public_key and provider_public_key for start.

PR #151, opened 2018-02-16
Eth address in TaskToCompute

jiivan:
fixes: #150

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -1,6 +1,8 @@
 import enum
 import functools
 
+from ethereum.utils import sha3
+
 from golem_messages import datastructures
 from golem_messages import exceptions
 from golem_messages import validators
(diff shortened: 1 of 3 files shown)


tests/message/test_tasks.py
```diff
@@ -115,3 +121,13 @@ def test_concent_enabled_default_true(self):
...
+
+    def test_ethereum_address(self):
+        msg = factories.TaskToComputeFactory()
+        serialized = shortcuts.dump(msg, None, None)
+        msg_l = shortcuts.load(serialized, None, None)
+        for addr_slot in (
+                'requestor_ethereum_address',
+                'provider_ethereum_address'):
+            address = getattr(msg_l, addr_slot)
+            self.assertEqual(len(address), 2 + (20*2))
```
rwrzesien: Wouldn't it be worth to compare `requestor_ethereum_public_key` and `provider_ethereum_public_key` before and after serialization ?

jiivan: Actually it's already done in `SerializationMixin`.

golem_messages/message/tasks.py
```diff
@@ -89,6 +91,14 @@ def __init__(self, header: datastructures.MessageHeader = None,
         if self.concent_enabled is None:
             self.concent_enabled = True
 
+    @property
+    def requestor_ethereum_address(self):
+        return '0x{}'.format(self.requestor_ethereum_public_key[:20].hex())
```
Krigpl: This isn't correct. The right formula is `address = sha3(public_key)[12:]`. See `pubkeytoaddr` in `golem.utils`

Krigpl:
Actually, it is really needed? Since we can easily compute it from the public key, so we can pass only the public key.

jiivan:
It's contrary to the documentation, so I believe it will prove useful.

PR #160, opened 2018-02-22
Add RemoveTaskContainer message

badb:
Alternative implemenation of: https://github.com/golemfactory/golem-messages/pull/149

Please choose the one that you prefer. I personally vote for this one, but I'm not 100% convince that's the right approach.

### DIFF
--- a/golem_messages/message/p2p.py
+++ b/golem_messages/message/p2p.py
@@ -149,14 +149,14 @@ class SetTaskSession(base.Message):
 
 class RemoveTaskContainer(base.Message):
     """
-    Message that contains RemoveTask message signed by task owner. Other nodes
-    uses this container to propagate information about task cancellation.
+    Message that contains RemoveTask messages signed by task owners. Other nodes
+    uses this container to propagate information about tasks cancellation.
     """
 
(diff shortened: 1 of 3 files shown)


golem_messages/message/p2p.py
```diff
@@ -141,3 +141,13 @@ class SetTaskSession(base.Message):
         'conn_id',
         'super_node_info',
     ] + base.Message.__slots__
+
+
+class RemoveTaskContainer(base.Message):
+    TYPE = P2P_MESSAGE_BASE + 17
+
+    __slots__ = ['remove_task'] + base.Message.__slots__
```
jiivan: Could we make it `remove_tasks` list? Even if for now we will put only single message there.

golem_messages/message/p2p.py
```diff
@@ -141,3 +141,13 @@ class SetTaskSession(base.Message):
         'conn_id',
         'super_node_info',
     ] + base.Message.__slots__
+
+
+class RemoveTaskContainer(base.Message):
```
jiivan: A docstring commenting why and how it's used would be helpful.

PR #163, opened 2018-02-28
Fixed equality method

jiivan:


tests/test_basic.py
```diff
@@ -32,6 +32,24 @@ def test_total_basic(self):
                                    self.ecc.raw_pubkey)
         self.assertEqual(msg, msg2)
 
+    def test_inequal_slots(self):
+        msg1 = message.RandVal(rand_val=1)
+        msg2 = message.RandVal(rand_val=2)
+        self.assertNotEqual(msg1, msg2)
+
+    def test_inequal_header(self):
```
shadeofblue: @jiivan shouldn't there be a positive test for equality here? or are we counting `test_total_basic` as one?

tests/test_basic.py
```diff
@@ -134,6 +152,12 @@ def test_hello_version_signature(self):
         with self.assertRaises(exceptions.InvalidSignature):
             self.ecc.verify(msg.sig, msg.get_short_hash())
 
+    def test_hello_version_inequality(self):
```
shadeofblue: @jiivan and also here? shouldn't there be a positive equality test?

PR #164, opened 2018-02-28
Don't duplicate in ComputeTaskDef parameters from TaskHeader

Wiezzel:
Removed fields:
  * `return_address`
  * `return_port`
  * `task_owner`
  * `key_id`
  * `environment`

Related issue: https://github.com/golemfactory/golem/issues/2073

### DIFF
--- a/tests/message/test_tasks.py
+++ b/tests/message/test_tasks.py
@@ -89,27 +89,6 @@ class TaskToComputeTest(unittest.TestCase,
         msg = shortcuts.load(serialized, None, None)
         self.assertIsInstance(msg, message.tasks.TaskToCompute)
 
-    def test_task_to_compute_validate_compute_task_def(self):
-        requestor_id = 'such as epidemiology'
-        # Shoudn't raise
-        message.TaskToCompute(slots=(
-            ('requestor_id', requestor_id),
-        ))
-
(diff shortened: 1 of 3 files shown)


tests/message/test_tasks.py
```diff
@@ -89,27 +89,6 @@ def test_task_to_compute_basic(self):
         msg = shortcuts.load(serialized, None, None)
         self.assertIsInstance(msg, message.tasks.TaskToCompute)
 
-    def test_task_to_compute_validate_compute_task_def(self):
```
shadeofblue: @Wiezzel well, I wouldn't remove the _whole_ test here - just remove the last check that tries to force the requestor_id to be different than the task owner...

Wiezzel: @shadeofblue Well, I removed the _whole_ method, so I don't see any reason for keeping this test case. What would it be?

shadeofblue: @Wiezzel okay, I was under impression that this test does a bit more than just test that single method but you may be right that that's not the case... though, slight decrease of coverage would suggest that it indeed does... but yes, that might not be an issue here...

badb:
I like this change, but we should confirm with @lukasz-glen that it's not in big conflict with RDProjekt changes (it shouldn't but still).

PR #173, opened 2018-03-15
make the `FileTranferToken.operation` an Enum field

shadeofblue:


golem_messages/message/concents.py
```diff
@@ -152,6 +152,15 @@ class FileTransferToken(base.Message):
     """
     TYPE = CONCENT_MSG_BASE + 5
 
+    @enum.unique
```
jiivan: `StringEnum`?

shadeofblue: ok

PR #177, opened 2018-03-23
remove unused DeltaParts message

etam:
Depends on https://github.com/golemfactory/golem/pull/2479


shadeofblue:
@etam @jiivan  hmm... we should first make some decisions re: code-freeze et al

etam:
Maybe this repo should also have a branch for code freeze? Then such could be merged into master, without worrying about stable branch.

PR #179, opened 2018-03-27
Add subtask_price to TaskToCompute message

### LINKED ISSUE #157: Add subtask_price to the message TaskToCompute
And a proper comment like "total subtask price computed as price*subtask_timesout". And maybe some validation too, like subtask_price > 0.

kbeker:
Resolves #157

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -96,7 +96,7 @@ class TaskToCompute(TaskMessageMixin, base.Message):
         'compute_task_def',
         'package_hash',
         'concent_enabled',
-        'price', # total subtask price computed as price*subtask_timesout
+        'price', # total subtask price computed as `price * subtask_timeout`
     ] + base.Message.__slots__
 
     def __init__(self, header: datastructures.MessageHeader = None,
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -96,6 +96,7 @@ class TaskToCompute(TaskMessageMixin, base.Message):
         'compute_task_def',
         'package_hash',
         'concent_enabled',
+        'price', # total subtask price computed as price*subtask_timesout
     ] + base.Message.__slots__
 
     def __init__(self, header: datastructures.MessageHeader = None,
@@ -123,6 +124,11 @@ class TaskToCompute(TaskMessageMixin, base.Message):
         value = super().deserialize_slot(key, value)
         if key == 'compute_task_def':
             value = ComputeTaskDef(value)
+        if key == 'price':
+            validators.validate_integer(
+                field_name='price',
+                value=value,
+            )
         return value
 
     @property
--- a/golem_messages/validators.py
+++ b/golem_messages/validators.py
@@ -10,3 +10,12 @@ def validate_varchar(field_name, value, max_length):
             field=field_name,
             value=value,
         )
+
+
+def validate_integer(field_name, value):
+    if not isinstance(value, int):
+        raise exceptions.FieldError(
+            "Should be an integer field",
+            field=field_name,
+            value=value,
+        )
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -30,3 +30,21 @@ class ValidateVarcharTestCase(unittest.TestCase):
     def test_value_invalid_type(self):
         with self.assertRaises(exceptions.FieldError):
             self.validate(fake.pyint())
+
+
+class ValidateIntegerTestCase(unittest.TestCase):
+    def setUp(self):
+        self.field_name = fake.word()
+
+    def validate(self, value):
+        return validators.validate_integer(
+            field_name=self.field_name,
+            value=value,
+        )
+
+    def test_value_valid(self):
+        self.validate(fake.pyint())
+
+    def test_value_invalid_type(self):
+        with self.assertRaises(exceptions.FieldError):
+            self.validate(fake.word())
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -50,6 +50,8 @@ class TaskToComputeFactory(factory.Factory):
 
     compute_task_def = factory.SubFactory(ComputeTaskDefFactory)
 
+    price = factory.Faker('random_int', min=1 << 20, max=10 << 20)
+
 
 class SubtaskResultsAcceptedFactory(factory.Factory):
     class Meta:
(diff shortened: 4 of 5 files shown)


golem_messages/message/tasks.py
```diff
@@ -107,6 +108,9 @@ def __init__(self, header: datastructures.MessageHeader = None,
         if self.concent_enabled is None:
             self.concent_enabled = True
 
+        if isinstance(self.subtask_price, int):
+            assert self.subtask_price > 0
```
shadeofblue: @kbeker is this requirement specified anywhere in the documentation? @lukasz-glen perhaps?

golem_messages/message/tasks.py
```diff
@@ -96,6 +96,7 @@ class TaskToCompute(TaskMessageMixin, base.Message):
         'compute_task_def',
         'package_hash',
         'concent_enabled',
+        'subtask_price', # total subtask price computed as price*subtask_timesout
```
shadeofblue: typo -> should be `subtask_timeout` + I'd call it just `price`

golem_messages/message/tasks.py
shadeofblue: @kbeker ah, just noticed @lukasz-glen 's comment in the issue

jiivan: 1. Probably `deserialize_slot()` is a better place for verification.
1. Please use `FieldError` ( see: https://github.com/golemfactory/golem-messages/blob/9edd8179f2c9a9aedc00879f505e056dddd9631a/golem_messages/validators.py#L6-L12 )
1. Assertions are disabled in optimized mode and shouldn't be an integral part of production logic: https://docs.python.org/3/library/constants.html#__debug__

golem_messages/message/tasks.py
kbeker: @shadeofblue, Łukasz wrote on issue 
`And a proper comment like "total subtask price computed as price*subtask_timesout". And maybe some validation too, like subtask_price > 0.`
Shouldn't be `subtask_price` as he said?

golem_messages/message/tasks.py
kbeker: Done, @jiivan please look at it one more time :)

tests/test_validators.py
```diff
@@ -19,14 +20,23 @@ def validate(self, value):
             max_length=self.max_length,
         )
 
+    def validate_subtask_price_value(self, value):
+        return validators.validate_integer(
+            field_name=self.field_name,
+            value=value,
+        )
```
jiivan: I believe this is unrelated to `ValidateVarcharTestCase`

golem_messages/validators.py
```diff
@@ -10,3 +10,12 @@ def validate_varchar(field_name, value, max_length):
             field=field_name,
             value=value,
         )
+
+
+def validate_integer(field_name, value):
+    if not ((isinstance(value, int)) or value is None):
```
jiivan: Can `value` be `None` and still an instance of 'int'?

golem_messages/validators.py
```diff
@@ -10,3 +10,12 @@ def validate_varchar(field_name, value, max_length):
...
+        raise exceptions.FieldError(
+            "Should be a integer field",
```
jiivan: `an integer`

golem_messages/message/tasks.py
shadeofblue: @kbeker nah, just make it `price` so as not to cause further proliferation of unneeded `subtask_` vs `task_` prefixes...

tests/test_validators.py
shadeofblue: indeed, @kbeker please create a separate test case

tests/test_validators.py
```diff
@@ -19,14 +20,23 @@ def validate(self, value):
...
+
     def test_value_too_long(self):
         with self.assertRaises(exceptions.FieldError):
             self.validate('*'*(self.max_length+1))
 
     def test_value_valid(self):
         for x in range(self.max_length+1):
             self.validate('*'*x)
+        self.validate_subtask_price_value(fake.pyint())
```
shadeofblue: this should be a separate test method ... though, generally, as mentioned above, it should be a separate test case anyway...

golem_messages/message/tasks.py
shadeofblue: thank you

golem_messages/message/tasks.py
```diff
@@ -123,6 +125,11 @@ def deserialize_slot(self, key, value):
         value = super().deserialize_slot(key, value)
         if key == 'compute_task_def':
             value = ComputeTaskDef(value)
+        if key == 'subtask_price':
```
shadeofblue: + please add two tests that load `TaskToCompute` with an integer, and a non-integer value in this field... as, sure, a test of the validator is fine, but I was more concerned about the `TaskToCompute` message itself...

kbeker:
I have added all in one fixUp, @jiivan, @shadeofblue

shadeofblue:
looks good now, imvho, though, I'd wait for @jiivan to have a look too

plus @lukasz-glen I'm unsure how all this relates to today's discussion re: price calculation in golem core (on #devs) -> maybe we should make sure once more that what we want to do here is indeed what should be done

golem_messages/message/tasks.py
```diff
@@ -107,6 +108,10 @@ def __init__(self, header: datastructures.MessageHeader = None,
         if self.concent_enabled is None:
             self.concent_enabled = True
 
+        if self.price is None:
+            self.price = 0
```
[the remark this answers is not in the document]
kbeker: @jiivan Better to leave it None value?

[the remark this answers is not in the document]
jiivan: Yes, I believe so. `None` is default expected behaviour. `0` is custom.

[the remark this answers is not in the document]
kbeker: Ok. I've added to it also to validation that it could be None value.

golem_messages/validators.py
```diff
@@ -13,7 +13,7 @@ def validate_varchar(field_name, value, max_length):
 
 
 def validate_integer(field_name, value):
-    if not isinstance(value, int):
+    if not isinstance(value, int) and value is not None:
```
shadeofblue: @kbeker I'd just leave the previous check in place and treat `None` as illegal ... it should be the responsibility of the code that creates this message to ensure that it always sets the price... 

btw, that's e.g. why I had suggested, in one of my previous comments (https://github.com/golemfactory/golem-messages/pull/179#issuecomment-376936222) to add this property to the factory (here ->
https://github.com/golemfactory/golem-messages/blob/master/tests/factories.py#L36) so any tests that rely on factory-provided `TaskToCompute`, receive a well-formed message...

kbeker: OK. I've added all, also I had to add few lines to `test_basic.py`

PR #187, opened 2018-04-09
remove `Message.header.encrypted` from the message equality criteria

shadeofblue:


golem_messages/message/base.py
```diff
@@ -241,7 +241,8 @@ def __eq__(self, obj):
         return (
```
jiivan: Can we add comment/docstring here to explicitly state that we're ignoring `.encrypted` in comparison?

shadeofblue: @jiivan sure! :)

PR #193, opened 2018-04-12
fix factories

shadeofblue:
* fix `VerdictReportComputedTaskFactory` 
* fix `ForceReportComputedTaskResponseFactory`

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -13,6 +13,8 @@ from golem_messages import message
 from golem_messages import serializer
 from golem_messages import datastructures
 
+from golem_messages.factories.helpers import clone_message
+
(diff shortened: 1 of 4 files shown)


tests/test_basic.py
```diff
@@ -30,15 +32,6 @@ class MessageEqualityTest(unittest.TestCase):
...
         self.ecc = golem_messages.ECCx(None)
 
-    @staticmethod
-    def _clone_message(msg, override_class=None):
-        msg_class = override_class or msg.__class__
-        return msg_class(
-            header=msg.header,
-            sig=msg.sig,
-            slots=msg.slots(),
-        )
```
jiivan: It always ends like this. There is no place for staticmethods :)

PR #201, opened 2018-04-20
Bugfix #194

### LINKED ISSUE #194: Semver support and bugfixes for version.py
[version.py](https://github.com/golemfactory/golem-messages/blob/master/version.py#L44) is based on [PEP 386](https://www.python.org/dev/peps/pep-0386/). In common cases like `0.5.3` it's compatible with [semver](https://semver.org) but more specific features like pre-release versions differ (`0.5.3-rc1` in semver vs `0.5.3rc1` in PEP 386). We want the script to fully support the semver form.

The problem is that `call_git_describe(prefix='v')` can't handle git tags like `v0.5.3-rc1`. It assumes that if there are hyphens in the output of `git describe`, they were added by git due to the commit not having a tag (e.g. `v0.5.2-8-gc3da727`).

1) Make the script handle versions like `0.5.3-rc1`.
    - Version may contain any number of hyphens. The presence of one or more hyphens should not be enough to assume that the commit is untagged. Instead of parsing the output of `git describe` in Python it might be better to use `git show` to print the tag and all the other info already in the desired format.
2) `RELEASE-VERSION` file is not updated when `call_git_describe()` fails.
    - `call_git_describe()` returns `None` when it fails for any reason. But this does not mean that `RELEASE-VERSION` file does not exist. It may exist and contain a valid version.
    - When that happens, `get_version()` returns the new version but **does not** update `RELEASE-VERSION` because it thinks the version did not change.
3) Make error handling in the script more robust.
    - The script catches `Exception` in multiple places and this way many problems that should be handled differently are treated the same way - silenced.
    - This also obscures the original cause of a problem. I think that the script should catch only the specific exception types it knows what to do with. Everything else should be let through and crash it so that it's noticed quickly and fixed.

kbeker:
Resolves #194

### DIFF
--- a/version.py
+++ b/version.py
@@ -40,18 +40,27 @@ VERSION_FILE = "RELEASE-VERSION"
 
 
 def call_git_describe(prefix='', cwd='.'):
-    cmd = 'git describe --tags --match %s[0-9]*' % prefix
+    version_name_cmd = 'git describe --tags --match %s[0-9]*' % prefix
+    last_tag_cmd = 'git describe --tags --abbrev=0'
     try:
         version = subprocess.run(
-            cmd.split(),
+            version_name_cmd.split(),
             stdout=subprocess.PIPE,
             check=True,
         ).stdout.decode()
+
+        last_tag = subprocess.run(
+            last_tag_cmd.split(),
+            stdout=subprocess.PIPE,
+            check=True,
+        ).stdout.decode()
+
         version = version.strip()[len(prefix):]
-        if '-' in version:
-            version = '{}+dev{}.{}'.format(*version.split('-'))
+        last_tag = last_tag.strip()[len(prefix):]
+        if version != last_tag:
+            version = f"{last_tag}+dev{version[len(last_tag)+1:].replace('-','.')}"
         return version
-    except Exception:
+    except subprocess.CalledProcessError:
         return None
 
 
@@ -60,7 +69,7 @@ def get_version(prefix='', cwd='.'):
     try:
         with path.open("r") as f:
             release_version = f.read()
-    except Exception:
+    except FileNotFoundError:
         release_version = None
 
     version = call_git_describe(prefix, cwd)
--- a/version.py
+++ b/version.py
@@ -34,9 +34,9 @@
 
 __all__ = ("get_version")
 
+import argparse
 import pathlib
 import subprocess
-import sys
 VERSION_FILE = "RELEASE-VERSION"
 
 
@@ -75,7 +75,7 @@ def call_git_describe(prefix='', cwd='.'):
     return version
 
 
-def get_version(prefix='', cwd='.', update_version_file=True):
+def get_version(prefix='', cwd='.', no_update_version_file=False):
     path = pathlib.Path(cwd) / VERSION_FILE
     try:
         with path.open("r") as f:
@@ -85,7 +85,7 @@ def get_version(prefix='', cwd='.', update_version_file=True):
 
     version = call_git_describe(prefix, cwd)
 
-    if not update_version_file:
+    if no_update_version_file:
         return version
 
     if version is None:
@@ -99,10 +99,7 @@ def get_version(prefix='', cwd='.', update_version_file=True):
 
 
 if __name__ == "__main__":
-    if '--no-update-version-file' in sys.argv and len(sys.argv) == 2:
-        update_version_file = False
-    elif len(sys.argv) == 1:
-        update_version_file = True
-    else:
-        raise ValueError('Invalid arguments')
-    print(get_version(prefix='v', update_version_file=update_version_file))
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--no-update-version-file', action='store_true', default=False)
+    args = parser.parse_args()
+    print(get_version(prefix='v', no_update_version_file=args.no_update_version_file))
--- a/version.py
+++ b/version.py
@@ -36,6 +36,7 @@ __all__ = ("get_version")
 
 import pathlib
 import subprocess
+import sys
 VERSION_FILE = "RELEASE-VERSION"
 
 
@@ -57,21 +58,24 @@ def call_git_describe(prefix='', cwd='.'):
     ).stdout.decode()
 
     assert ' ' not in prefix
+    assert version.strip().startswith(prefix)
+    assert last_tag.strip().startswith(prefix)
+
     version = version.strip()[len(prefix):]
     last_tag = last_tag.strip()[len(prefix):]
+
+    assert version.startswith(last_tag)
     if version != last_tag:
-        """
-        Helper to generate version compatible with semver
-        Compares the last tag version name with the version name
-        where the repository is currently located
-        If it is not currently on tag adds '+dev' beetwen last tag name
-        and a number of commits behind it separated by hyphens with short commit hash
-        """
+        # `version` does not match the last tag so we must be on an untagged commit.
+        # In that case `version` is that tag with a suffix consisting of the number of commits and a commit ID.
+        # To make this very clear, we want to insert `+dev` between the tag and the suffix. Also, replace the
+        # hyphen in the suffix with a dot but without mangling any hyphens that might be in the tag.
+        # In effect something like 0.5.3-rc5-post1-5-78df3e12 becomes 0.5.3-rc5-post1+dev5.78df3e12.
         version = f"{last_tag}+dev{version[len(last_tag) + 1:].replace('-', '.')}"
     return version
 
 
-def get_version(prefix='', cwd='.', generate_version=True):
+def get_version(prefix='', cwd='.', update_version_file=True):
     path = pathlib.Path(cwd) / VERSION_FILE
     try:
         with path.open("r") as f:
@@ -79,13 +83,14 @@ def get_version(prefix='', cwd='.', generate_version=True):
     except FileNotFoundError:
         release_version = None
 
-    if generate_version:
-        version = call_git_describe(prefix, cwd)
-    else:
-        return release_version
+    if not update_version_file:
+        if not release_version:
+            raise ValueError("Cannot find the version number!")
+        else:
+            return release_version
 
-    if version is None:
-        version = release_version
+    version = call_git_describe(prefix, cwd)
+
     if version is None:
         raise ValueError("Cannot find the version number!")
 
@@ -97,4 +102,8 @@ def get_version(prefix='', cwd='.', generate_version=True):
 
 
 if __name__ == "__main__":
-    print(get_version(prefix='v'))
+    if '--no-update-version-file' in sys.argv:
+        update_version_file = False
+    else:
+        update_version_file = True
+    print(get_version(prefix='v', update_version_file=update_version_file))
--- a/version.py
+++ b/version.py
@@ -83,13 +83,10 @@ def get_version(prefix='', cwd='.', update_version_file=True):
     except FileNotFoundError:
         release_version = None
 
-    if not update_version_file:
-        if not release_version:
-            raise ValueError("Cannot find the version number!")
-        else:
-            return release_version
-
     version = call_git_describe(prefix, cwd)
+
+    if not update_version_file:
+        return version
 
     if version is None:
         raise ValueError("Cannot find the version number!")
@@ -105,5 +102,5 @@ if __name__ == "__main__":
     if '--no-update-version-file' in sys.argv:
         update_version_file = False
     else:
-        update_version_file = True
+        raise ValueError('Cannot get appropriate parameter')
     print(get_version(prefix='v', update_version_file=update_version_file))
--- a/version.py
+++ b/version.py
@@ -40,31 +40,38 @@ VERSION_FILE = "RELEASE-VERSION"
 
 
 def call_git_describe(prefix='', cwd='.'):
-    version_name_cmd = 'git describe --tags --match %s[0-9]*' % prefix
-    last_tag_cmd = 'git describe --tags --abbrev=0'
-    try:
-        version = subprocess.run(
-            version_name_cmd.split(),
-            stdout=subprocess.PIPE,
-            check=True,
-        ).stdout.decode()
+    version_cmd = 'git describe --tags --match %s[0-9]*' % prefix
+    last_tag_cmd = 'git describe --tags --match %s[0-9]* --abbrev=0' % prefix
+    version = subprocess.run(
+        version_cmd.split(),
+        stdout=subprocess.PIPE,
+        check=True,
+        shell=False,
+    ).stdout.decode()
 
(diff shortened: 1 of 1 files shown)


version.py
```diff
@@ -60,7 +69,7 @@ def get_version(prefix='', cwd='.'):
     try:
         with path.open("r") as f:
             release_version = f.read()
-    except Exception:
+    except FileNotFoundError:
```
jiivan: :+1:

version.py
```diff
@@ -40,18 +40,27 @@
...
+
         version = version.strip()[len(prefix):]
-        if '-' in version:
-            version = '{}+dev{}.{}'.format(*version.split('-'))
+        last_tag = last_tag.strip()[len(prefix):]
+        if version != last_tag:
+            version = f"{last_tag}+dev{version[len(last_tag)+1:].replace('-','.')}"
         return version
-    except Exception:
+    except subprocess.CalledProcessError:
```
jiivan: :+1:

version.py
```diff
@@ -40,18 +40,27 @@
...
+            stdout=subprocess.PIPE,
+            check=True,
+        ).stdout.decode()
+
         version = version.strip()[len(prefix):]
-        if '-' in version:
-            version = '{}+dev{}.{}'.format(*version.split('-'))
+        last_tag = last_tag.strip()[len(prefix):]
+        if version != last_tag:
+            version = f"{last_tag}+dev{version[len(last_tag)+1:].replace('-','.')}"
```
jiivan: Looks good but I would like to see some comment why it's done this way. I.e. that it helps with tags that have hyphens in them like described in an issue.

version.py
cameel: I think that we should not catch `CalledProcessError`. This error means that the command failed, e.g. because the process has returned a non-zero exit code. I'd prefer the script to fail immediately and let me know that there's a problem rather than return `None` and proceed.

Also, as long as this function can return `None`, point 2 from #194 is not addressed.

version.py
```diff
@@ -40,18 +40,27 @@
...
+        last_tag = subprocess.run(
+            last_tag_cmd.split(),
+            stdout=subprocess.PIPE,
+            check=True,
+        ).stdout.decode()
+
         version = version.strip()[len(prefix):]
-        if '-' in version:
-            version = '{}+dev{}.{}'.format(*version.split('-'))
+        last_tag = last_tag.strip()[len(prefix):]
```
cameel: This won't work correctly if `prefix` contains leading whitespace. I'd add an `assert` against that.

Actually, given that `prefix` gets embedded in a bash command above, the function won't work if it contains any whitespace - leading, trailing or spaces between words.

version.py
```diff
@@ -40,18 +40,27 @@
...
-            cmd.split(),
+            version_name_cmd.split(),
             stdout=subprocess.PIPE,
             check=True,
         ).stdout.decode()
+
+        last_tag = subprocess.run(
+            last_tag_cmd.split(),
+            stdout=subprocess.PIPE,
+            check=True,
```
cameel: We're not relying on any shell features. It would be better to run both commands with `shell=False`.

version.py
```diff
@@ -40,18 +40,27 @@
 
 
 def call_git_describe(prefix='', cwd='.'):
-    cmd = 'git describe --tags --match %s[0-9]*' % prefix
+    version_name_cmd = 'git describe --tags --match %s[0-9]*' % prefix
+    last_tag_cmd = 'git describe --tags --abbrev=0'
```
cameel: Isn't `--match` necessary in both commands? Otherwise `last_tag_cmd` can be a tag that does not have the prefix. Below you're assuming that it must have the prefix because you're stripping it.

Since it's obviously your assumption that both `version` and `last_tag` start with the prefix and the code is broken if they don't maybe it would be good idea to have assertions against that. If you had these asserts they would catch this problem.

cameel: Why not `version_cmd` for consistency with `version` variable below?

version.py
cameel: I see that there's an assumption that `version_name_cmd` is either equal to `last_tag_cmd` or contains it as a prefix. Adding an `assert` for this assumption would make that more apparent to someone reading the code and also, conveniently, let Python check and and alert us in development if your assumption does not hold true some case.

cameel: This style violates PEP8. You should add spaces around arithmetical operators and after commas:
``` python
version = f"{last_tag}+dev{version[len(last_tag) + 1:].replace('-', '.')}"
```

kbeker: I did this this way because we want to support all versioning so I needed to get last tag name, and name of tag where present code is and then see if there is any difference - if is add prefix `+dev` beetwen last tag name and number commits behind this tag.

I didn't know where you want comment - here - or in code so if you want this in code I will rewrite it. :)

version.py
kbeker: @cameel 
```call_git_describe() returns None when it fails for any reason. But this does not mean that RELEASE-VERSION file does not exist. It may exist and contain a valid version.```
There is actually a check that if there is any version in RELEASE-VERSION and function `call_git_describe()` returns None it won't change version what is in this file.
```   
    if version is None:
        version = release_version
    if version is None:
        raise ValueError("Cannot find the version number!")

    if version != release_version:
        with path.open("w") as f:
            f.write(version)
```

File: version.py (hunk 3 above)
Comment (kbeker) [reply version.py]: @jiivan Do you want some comment in code about how it's done or here?

File: version.py (hunk 4 above)
Comment (kbeker) [reply version.py]: Done

File: version.py (hunk 5 above)
Comment (kbeker) [reply version.py]: Done

File: version.py (hunk 6 above)
Comment (kbeker) [reply version.py]: Done

Comment (kbeker) [reply version.py]: Done

File: version.py (hunk 3 above)
Comment (kbeker) [reply version.py]: Done

Comment (kbeker) [reply version.py]: Also added asset so @cameel part Done

Comment (jiivan) [reply version.py]: In code

File: version.py (hunk 2 above)
Comment (cameel) [reply version.py]: But that's what I mean. If this happens, `RELEASE-VERSION` is not updated but the script does not fail. Which means that the caller will think that everything went fine and use the wrong version.

That's something that actually happened for us when we encountered the problem with dashes in version numbers. @bartoszbetka reported that the `RELEASE-VERSION` is not always updated when we run the script to generate it, even though it's supposed to be.

The function should fail if this happens.

File: version.py (hunk 5 above)
Comment (cameel) [reply version.py]: Did you push all your changes? I don't see `shell=False` in the code.

File: version.py [hunk 7]
```diff
@@ -40,18 +40,28 @@
...
+            check=False,
+        ).stdout.decode()
+
+        assert ' ' not in prefix
         version = version.strip()[len(prefix):]
-        if '-' in version:
-            version = '{}+dev{}.{}'.format(*version.split('-'))
+        last_tag = last_tag.strip()[len(prefix):]
+        if version != last_tag:
+            version = f"{last_tag}+dev{version[len(last_tag) + 1:].replace('-', '.')}"
```
cameel: Please replace `+dev` with `-dev` according to our discussion on Rocket.

Of course do it in a separate (non-fixup) commit since this does significantly change the output so it should be visible as a separate change.

cameel: What about https://github.com/golemfactory/golem-messages/pull/201#discussion_r183093950?

version.py
kbeker: @cameel 
Yes, I have changed it and I pushed everything. See: https://github.com/golemfactory/golem-messages/pull/201/files#diff-78c8de9f11870b74f87e1deeffeccb84R49

kbeker: I have read that you want me to change `check`. Already fixed. Done

version.py
kbeker: @cameel So the best way to do this will be deleting try-except part of code and if only any error will appear it will fail. 
I change it as I wrote. Check if this is ok

version.py
kbeker: Done

version.py
kbeker: Done

kbeker: @cameel see my qustion on Rocket please

version.py
cameel: Yeah, but please make sure you test it in all relevant scenarios:
- reading `RELEASE-VERSION` or generating it
- `git describe` fails or does not fail (e.g. because git is not installed)
- `RELEASE-VERSION` exists on disk or does not exist.

Please check all 8 combinations. I think removing this handler may have broken reading `RELEASE-VERSION` when git is not installed. If that's the case, more changes would be needed.

version.py
cameel: The consensus seems to be that the fix should be elsewhere. So let's leave it at `+dev`.

version.py
kbeker: @cameel 
Done. Now if we use function `get_version(generate_version=False)` we only read version from file exactly as you said.

cameel: So please remove the exception handlers from `call_git_describe()` now.

kbeker: Done, exception handlers removed

version.py
```diff
@@ -40,30 +40,49 @@
...
             release_version = f.read()
-    except Exception:
+    except FileNotFoundError:
         release_version = None
 
-    version = call_git_describe(prefix, cwd)
+    if generate_version:
+        version = call_git_describe(prefix, cwd)
+    else:
+        return release_version
```
cameel: When `generate_version` is `False` and `RELEASE-VERSION` does not exist the script should fail.

You can do this by raising an exception there or later, inside the `__main__` block.

version.py
```diff
@@ -40,30 +40,49 @@
...
 
     if version is None:
         version = release_version
```
cameel: This part can be simplified now because `version` will never be `None`.

version.py
```diff
@@ -40,30 +40,49 @@
...
+
+    assert ' ' not in prefix
+    version = version.strip()[len(prefix):]
+    last_tag = last_tag.strip()[len(prefix):]
+    if version != last_tag:
+        """
+        Helper to generate version compatible with semver
+        Compares the last tag version name with the version name
+        where the repository is currently located
+        If it is not currently on tag adds '+dev' beetwen last tag name
```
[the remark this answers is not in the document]
kbeker: Done

version.py
```diff
@@ -40,30 +40,49 @@
...
+    ).stdout.decode()
+
+    assert ' ' not in prefix
+    version = version.strip()[len(prefix):]
+    last_tag = last_tag.strip()[len(prefix):]
+    if version != last_tag:
+        """
+        Helper to generate version compatible with semver
+        Compares the last tag version name with the version name
+        where the repository is currently located
```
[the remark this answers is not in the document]
kbeker: Done

version.py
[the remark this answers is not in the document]
kbeker: Done

version.py
```diff
@@ -40,30 +40,49 @@
...
+        If it is not currently on tag adds '+dev' beetwen last tag name
+        and a number of commits behind it separated by hyphens with short commit hash
+        """
+        version = f"{last_tag}+dev{version[len(last_tag) + 1:].replace('-', '.')}"
```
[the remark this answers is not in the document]
kbeker: Done

version.py
```diff
@@ -40,30 +40,49 @@
...
+
+    last_tag = subprocess.run(
+        last_tag_cmd.split(),
+        stdout=subprocess.PIPE,
+        check=True,
+        shell=False,
+    ).stdout.decode()
+
+    assert ' ' not in prefix
+    version = version.strip()[len(prefix):]
```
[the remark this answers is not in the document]
kbeker: Done

version.py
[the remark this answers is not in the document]
kbeker: Done

version.py
kbeker: Done

version.py
kbeker: Do you want me to do specific exception or just `Exception`?

kbeker:
@cameel 
All changes added.

version.py
cameel: I would personally always use a custom exception class but since the code is already using `ValueError` raising it is fine too.

kbeker: So it is done

version.py
```diff
@@ -36,37 +36,61 @@
...
     path = pathlib.Path(cwd) / VERSION_FILE
     try:
         with path.open("r") as f:
             release_version = f.read()
-    except Exception:
+    except FileNotFoundError:
         release_version = None
 
+    if not update_version_file:
+        if not release_version:
```
cameel: This should be `if release_version is None:`. Or `if release_version in [None, '']:` if you really wanted to raise the error when the file is empty.

version.py
```diff
@@ -78,4 +102,8 @@ def get_version(prefix='', cwd='.'):
 
 
 if __name__ == "__main__":
-    print(get_version(prefix='v'))
+    if '--no-update-version-file' in sys.argv:
+        update_version_file = False
+    else:
+        update_version_file = True
```
cameel: You missed this part:

>  It should fail if it gets anything else than this option or no options at all.

cameel: This could be shortened to
``` python
update_version_file = ('--no-update-version-file' in sys.argv')
```

kbeker: Done

kbeker: Good to know:) Done
@cameel ready for review

version.py
```diff
@@ -78,4 +99,10 @@ def get_version(prefix='', cwd='.'):
 
 
 if __name__ == "__main__":
-    print(get_version(prefix='v'))
+    if '--no-update-version-file' in sys.argv and len(sys.argv) == 2:
+        update_version_file = False
+    elif len(sys.argv) == 1:
+        update_version_file = True
+    else:
+        raise ValueError('Cannot get appropriate parameter')
```
cameel: Please reword: `Invalid arguments`

kbeker:
@jiivan Could you look at this PR now? Because @cameel already accepted it :) 
I don't know what is going on with `ci/circleci — Your tests failed on CircleCI` becuase I run tests locally and everything is fine. Any hints?

version.py
kbeker: Done :)

version.py
```diff
@@ -78,4 +99,10 @@ def get_version(prefix='', cwd='.'):
 
 
 if __name__ == "__main__":
-    print(get_version(prefix='v'))
+    if '--no-update-version-file' in sys.argv and len(sys.argv) == 2:
+        update_version_file = False
+    elif len(sys.argv) == 1:
+        update_version_file = True
+    else:
+        raise ValueError('Invalid arguments')
```
jiivan: We should probably use [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) instead of mocking our own solution from scratch.

kbeker: @jiivan Done, if it is OK, just merge it please.

Issue #196, opened 2018-04-16
Add Concent constants to golem-messages


shadeofblue:

cameel:
Constants currently defined in Concent:

``` python
CONCENT_MESSAGING_TIME    = 3600
FORCE_ACCEPTANCE_TIME     = 3600 * 4
MAXIMUM_DOWNLOAD_TIME     = 1800
SUBTASK_VERIFICATION_TIME = 3600 * 4
PAYMENT_DUE_TIME          = 3600 * 12
```

Note that [maximum_download_time()](https://github.com/golemfactory/golem-messages/blob/v2.3.4/golem_messages/helpers.py#L23-L39) is a constant in Concent but not in golem-messages. We're going to change that in Concent.

PR #202, opened 2018-04-23
Add concent time constants & helpers

jiivan:
fixes: #196 

concent constants: https://github.com/golemfactory/concent/blob/415d34f326b48fc964f0c3cd684e685cad042c0e/concent_api/concent_api/settings/base.py#L222-L246

related to: https://github.com/golemfactory/concent/issues/312

### DIFF
--- a/golem_messages/helpers.py
+++ b/golem_messages/helpers.py
@@ -4,6 +4,18 @@ import math
 from golem_messages import constants
 
 
+def subtask_verification_time(report_computed_task) -> datetime.timdelta:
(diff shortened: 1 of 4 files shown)


golem_messages/helpers.py
```diff
@@ -4,6 +4,18 @@
 from golem_messages import constants
 
 
+def subtask_verification_time(report_computed_task) -> datetime.timdelta:
```
shadeofblue: typo in the type -> `datetime.timedelta` ... does it not trigger some error?

PR #215, opened 2018-05-10
concent_enabled flag in WantToComputeTask

jiivan:
This allows Requestor to avoid sending `TaskToCompute` if it cannot meet Providers` requirements.


tests/message/test_messages.py
```diff
@@ -50,14 +50,16 @@ def test_message_want_to_compute_task(self):
         max_resource_size = random.randint(1, 2**10)
         max_memory_size = random.randint(1, 2**10)
         num_cores = random.randint(1, 2**5)
+        concent_enabled = random.random() > 0.5
```
shadeofblue: actually, I'm unsure if it makes sense to test with a random value like that ... I mean, imagine it fails with one value but not the other - this will make it harder to test if the tests are non-deterministic...

PR #222, opened 2018-05-17
Add `category` enum to `FileTransferToken.FileInfo`

### LINKED ISSUE #214: Add `type` enum to `FileTransferToken.FileInfo`
The `type` is needed to allow both parties to know what to expect within each of the files so that they don't need to rely on either the order of the files in the token or on the specific naming of those files

shadeofblue:
+ frozendict default values test
* pull `SetItemDict` out of `FrozenDict`
* ensure that `FrozenDict` defaults are really its defaults and not just `defaultdict`-like fallbacks
+ `ValidatingDict` class to abstract the validation interface from `ComputeTaskDef`
+ validation for `FileTransferToken.FileInfo`'s new `category` field
+ tests

closes #214

### DIFF
--- a/golem_messages/message/concents.py
+++ b/golem_messages/message/concents.py
@@ -132,18 +132,40 @@ class FileTransferToken(base.Message):
     ] + base.Message.__slots__
 
     def deserialize_slot(self, key, value):
+        def deserialize_fileinfo(f):
+            cat = f.get('category')
+            if cat in [c.value for c in FileTransferToken.FileInfo.Category]:
+                f['category'] = FileTransferToken.FileInfo.Category(cat)
+            return FileTransferToken.FileInfo(f)
+
         value = super().deserialize_slot(key, value)
         if key == 'files':
-            value = [FileTransferToken.FileInfo(f) for f in value]
+            value = [deserialize_fileinfo(f) for f in value]
         return value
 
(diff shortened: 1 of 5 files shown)


golem_messages/message/concents.py
```diff
@@ -132,19 +132,41 @@ class Operation(datastructures.StringEnum):
     ] + base.Message.__slots__
 
     def deserialize_slot(self, key, value):
+        def deserialize_fileinfo(f):
+            cat = f.get('category')
+            if cat in [c.value for c in FileTransferToken.FileInfo.Category]:
+                f['category'] = FileTransferToken.FileInfo.Category(cat)
```
jiivan: Maybe just:
```python
try:
    f['category'] = FileTransferToken.FileInfo.Category(cat)
except ValueError:
    pass
```

golem_messages/message/concents.py
```diff
@@ -153,6 +175,10 @@ def is_upload(self):
     def is_download(self):
         return self.operation == self.Operation.download
 
+    def get_file_info(self, category: FileInfo.Category):
+        fi = [fi for fi in self.files if fi.get('category', None) == category]
```
jiivan: overwritten `fi` label in list comprehension

PR #232, opened 2018-06-14
Add  new option -> ResultsFailed to SubtaskResultsSettled.Origin

dybi:


dybi:
Ok, it is not needed - we have misunderstood the docs ;)

PR #234, opened 2018-06-18
uuid/bytes32 conversion methods

Krigpl:
waffle


Krigpl:
@jiivan Can you help with `codecov`?

Issue #103, opened 2018-01-24
Remove registered_message_types


jiivan:
Remove `registered_message_types`. Assume all children of `golem_messages.message.base.Message` are valid messages unless they have `ABSTRACT` class attribute set to `True`

Another solution is to use library similar to the one in Django.

PR #242, opened 2018-07-23
Remove registered_message_types

jiivan:
fixes: #103


golem_messages/message/__init__.py
```diff
@@ -48,114 +59,4 @@
 from golem_messages.message.concents import ForceReportComputedTask  # noqa
```
shadeofblue: we might consider removing the concent-related messages from the list as they are probably imported using more specific imports...

Issue #241, opened 2018-07-19
`TaskToCompute.ethereum_address_proof`


shadeofblue:
* [x] add the field to TaskToCompute
* [x] add test(s)

PR #243, opened 2018-08-16
ethereum address proof

shadeofblue:
closes https://github.com/golemfactory/golem-messages/issues/241

### DIFF
--- a/golem_messages/message/base.py
+++ b/golem_messages/message/base.py
@@ -7,6 +7,7 @@ import logging
 import struct
 import time
 import warnings
+import typing
 
 import semantic_version
 
@@ -300,7 +301,8 @@ class Message():
         sha.update(payload or b'')
         return sha.digest()
 
-    def serialize(self, sign_as: bytes = None, encrypt_func=None):
+    def serialize(
+            self, sign_as: typing.Optional[bytes] = None, encrypt_func=None):
         """ Return serialized message
         :return str: serialized message """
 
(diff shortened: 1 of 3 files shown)


shadeofblue:
@jiivan I'd leave it as an excercise for a later time since it may not necessarily be trivial...

golem_messages/message/base.py
```diff
@@ -500,6 +500,24 @@ def valid_slot(self, name):
             and (name not in Message.__slots__) \
             and (name in self.__slots__)
 
+    def _verify_signature(
+            self, signature: bytes, public_key: bytes, msg_hash: bytes = None
```
Krigpl: `msg_hash: Optional[bytes] = None` imo

golem_messages/message/tasks.py
```diff
@@ -251,6 +254,8 @@ class TaskToCompute(ConcentEnabled, TaskMessage):
         'size',  # the size of the resources zip file
         'concent_enabled',
         'price',  # total subtask price computed as `price * subtask_timeout`
+
+        '_ethsig'  # must be last
```
Krigpl: Why is that? Is it because of some backward compatibility where it may or may not be set, thus having different length? It's going to always be set eventually, right? Why implement handling of the non set case?

shadeofblue: @Krigpl no, rather so that it's not treated as part of the message itself (in which case, the signature would be part of the payload it then signs)

PR #249, opened 2018-08-17
+ deprecation warning against the use of root `message` module imports

shadeofblue:


rwrzesien:
@shadeofblue Wasn't that already forced by `v2.11.0` update ?

PR #255, opened 2018-08-22
New task message: StateUpdate

inexxt:
This PR introduces new message type: `StateUpdate`. It is used to wrap Provider-Requestor communication during subtask execution. `data` field content is task-specific and golem is agnostic about it.

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -634,3 +634,28 @@ class RejectReportComputedTask(TaskMessage, base.AbstractReasonMessage):
     @base.verify_slot('cannot_compute_task', CannotComputeTask)
     def deserialize_slot(self, key, value):
         return super().deserialize_slot(key, value)
+
+@library.register(TASK_MSG_BASE + 31)
+class StateUpdate(base.Message):
+
+    @enum.unique
+    class DIRECTION(datastructures.StringEnum):
+        Call = enum.auto()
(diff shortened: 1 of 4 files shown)


golem_messages/message/tasks.py
```diff
@@ -634,3 +634,28 @@ class REASON(datastructures.StringEnum):
     @base.verify_slot('cannot_compute_task', CannotComputeTask)
     def deserialize_slot(self, key, value):
         return super().deserialize_slot(key, value)
+
+@library.register(TASK_MSG_BASE + 31)
+class StateUpdate(base.Message):
+
```
shadeofblue: @inexxt please add some docstring about the expected usage of this message

inexxt: Fixed

inexxt:
@shadeofblue Well, two message version is how it was done previously (in the previous attempt to bring Provider-Requestor communication to Golem) and then we decided it would be better to not introduce two different (but identical) messages, with the only difference being the direction.
This feature is meant to be very simple, for example to track progress of task execution. and I don't suspect there will be a need (in the foreseeable future) to differentiate StateUpdate-call and StateUpdate-response.
The code in main PR (https://github.com/golemfactory/golem/pull/3199/files) is already structured around  one-message version, and it would require rewriting quite a lot of it (+tests) for a questionable gain.

Issue #244, opened 2018-08-16
Provider has no way to ensure that their public key / address is indeed specified correctly in `TaskToCompute`


shadeofblue:
* Requestor can specify any other address and the "contract" between the provider and requestor (which `TaskToCompute` effectively constitues) still valid
* `TaskToCompute` should most likely contain the `WantToComputeTask` and provider data should be retrieved from this nested message and not be provided explicitly in TTC
* thus, `WantToComputeTask` should most likely include the `provider_public_key` and `provider_ethereum_public_key` fields and TTC should only have properties pointing to it
* following that, `ReportComputedTask` should not by itself contain the `eth_account` field and it should, again, be just a reference to the property in the TTC (which points to included WTCT)

shadeofblue:
FYI: @Krigpl @lukasz-glen @jiivan

PR #258, opened 2018-09-06
Feature provider has no way to ensure that their public key and address is indeed specified correctly in task to compute

rwrzesien:
Resolves https://github.com/golemfactory/golem-messages/issues/244.

### DIFF
--- a/golem_messages/factories/tasks.py
+++ b/golem_messages/factories/tasks.py
@@ -1,11 +1,9 @@
 # pylint: disable=too-few-public-methods,unnecessary-lambda
 import calendar
 import datetime
-import os
 import time
 import typing
 
-from eth_utils import encode_hex
 import factory.fuzzy
 import faker
 
@@ -27,6 +25,11 @@ class WantToComputeTaskFactory(helpers.MessageFactory):
 
     node_name = factory.Faker('name')
     task_id = factory.Faker('uuid4')
+    provider_public_key = factory.LazyFunction(
+        lambda: encode_key_id(cryptography.ECCx(None).raw_pubkey))
+    provider_ethereum_public_key = factory.SelfAttribute(
+        'provider_public_key'
+    )
 
 
 class CTDBlenderExtraDataFactory(factory.DictFactory):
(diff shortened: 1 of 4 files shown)


golem_messages/message/tasks.py
```diff
@@ -235,6 +235,9 @@ class WantToComputeTask(ConcentEnabled, base.Message):
         'price',
         'concent_enabled',  # Provider notifies requestor
                             # about his concent status
+        'provider_public_key',  # key used for msg signing and encryption
+        'provider_ethereum_public_key',  # used for transactions on blockchain
+        'eth_account',
```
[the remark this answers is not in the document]
rwrzesien: This was already there, and the issue description doesn't mention doing anything with it. Should it be removed or moved too ? If yes, please update issue description for consistency.

Btw. if this should be removed, please note that ie. Concent is using this field in few places.

golem_messages/message/tasks.py
```diff
@@ -416,6 +430,10 @@ class ReportComputedTask(TaskMessage):
         'options',
     ] + base.Message.__slots__
 
+    @property
+    def eth_account(self):
+        return self.task_to_compute.eth_account
```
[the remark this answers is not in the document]
rwrzesien: The issue description says:

> following that, ReportComputedTask should not by itself contain the eth_account field and it should, again, be just a reference to the property in the TTC (which points to included WTCT)

That is why it is there. If it changes somehow, please update the issue description.

golem_messages/message/tasks.py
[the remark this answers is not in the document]
jiivan: Please move `provider_ethereum_address` property to `WantToComputeTask`. There is no need to include both `provider_ethereum_public_key` and `eth_account` in one message. `eth_account` can be computed from `provider_ethereum_public_key`.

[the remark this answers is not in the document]
rwrzesien: Done, I have also move few tests which test this field on `TaskToCompute` to `WantToComputeTaskTest` test case.

tests/message/test_messages.py
```diff
@@ -51,6 +52,8 @@ def test_message_want_to_compute_task(self):
         max_memory_size = random.randint(1, 2**10)
         num_cores = random.randint(1, 2**5)
         concent_enabled = random.random() > 0.5
+        provider_public_key = cryptography.ECCx(None).raw_pubkey
```
shadeofblue: hmm.... why this particular value? the message normally contains:

```
encode_key_id(cryptography.ECCx(None).raw_pubkey))
```

so, a hex-encoded representation of the public key... 

so, if we want to stay true to the actually expected content of those fields, we should use the above for both `provider_public_key` and `provider_ethereum_public_key` ... and - as is the case here - if we don't care about the value being a representation of an actual public key, then we can use a random string for both ...

rwrzesien: @shadeofblue :+1: done, I think it is better to use real values when possible.

golem_messages/message/tasks.py
```diff
@@ -267,11 +274,18 @@ def requestor_ethereum_address(self):
...
-        )
+    def provider_public_key(self):
+        return self.want_to_compute_task.provider_public_key
+
+    @property
+    def provider_ethereum_public_key(self):
+        return self.want_to_compute_task.provider_ethereum_public_key
+
+    @property
+    def eth_account(self):
```
jiivan: Why change property name from `provider_ethereum_address` to `eth_account`?

rwrzesien: :+1: done, you are right, it is actually unnecessary change.

golem_messages/factories/tasks.py
```diff
@@ -63,19 +66,14 @@ class Meta:
...
     compute_task_def = factory.SubFactory(ComputeTaskDefFactory)
-    provider_public_key = factory.LazyFunction(
-        lambda: encode_key_id(cryptography.ECCx(None).raw_pubkey))
-    provider_ethereum_public_key = factory.SelfAttribute(
-        'provider_public_key'
-    )
     requestor_public_key = factory.LazyFunction(
         lambda: encode_key_id(cryptography.ECCx(None).raw_pubkey))
 
-    compute_task_def = factory.SubFactory(ComputeTaskDefFactory)
```
jiivan: :+1:

PR #264, opened 2018-09-17
add strict verification of slots (verify if the required slots are no…

### LINKED ISSUE #262: Change the `validate_slot` decorator to require/allow `None` values
add a `allow_none` parameter to the decorator so that the deserializer can check not only for the type of message but also require some values to be set

shadeofblue:
…t none)

closes #262

### DIFF
--- a/golem_messages/message/base.py
+++ b/golem_messages/message/base.py
@@ -77,8 +77,8 @@ def verify_version(msg_version):
         )
 
 
-def _verify_slot_type(value, class_):
-    if not isinstance(value, (class_, type(None))):
+def _verify_slot_type(value, class_, allow_none=False):
+    if not isinstance(value, (class_, type(None)) if allow_none else class_):
         raise TypeError(
             "Invalid nested message type {} should be {}".format(
                 type(value),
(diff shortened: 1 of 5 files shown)


golem_messages/message/base.py
```diff
@@ -77,8 +77,8 @@ def verify_version(msg_version):
         )
 
 
-def _verify_slot_type(value, class_):
-    if not isinstance(value, (class_, type(None))):
+def _verify_slot_type(value, class_, allow_none=False):
```
rwrzesien: How about naming it with the `is_` convention, as it is bool, ie. `is_none_allowed`? Just a proposal.

golem_messages/message/base.py
```diff
@@ -148,21 +151,23 @@ def _(self, key, value):
...
-def verify_slot_list(slot_name, item_class):
+def verify_slot_list(slot_name, item_class, allow_none=False):
     """
     decorator for Message's `deserialize_slot` method
     ensures that the slot identified by `slot_name` is a list of messages with
     the given instance type (provided in `item_class`)
 
     :param slot_name: the name of the slot to verify
     :param item_class: the class to check list items against
+    :param allow_none: whether we're allowing the slot to be empty
```
rwrzesien: I don't know if you do this or not, but maybe it is worth to add info about default value?

golem_messages/message/base.py
shadeofblue: I think `allow_none` is pretty obvious in this context ... and `is_none_allowed` seems like a name of a flag (property) rather than a parameter to a function...

PR #266, opened 2018-09-18
Feature replace script src field with dedicated classes for appropriate task

kbeker:
Resolves part of https://github.com/golemfactory/golem/issues/3066

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -96,8 +96,9 @@ class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
         'src_code': '',
         'script_template_hash': '',
         'task_type': '',
-        'meta_parameters': '',
-        'extra_data': {},  # safe because of copy in parent.__missing__()
+        'extra_data': {
+            'meta_parameters': {}
+        },  # safe because of copy in parent.__missing__()
         'short_description': '',
         'working_directory': '',
         'performance': 0,
@@ -120,34 +121,30 @@ class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
         max_length=128,
     )
 
-    def validate_meta_parameters(self, value):
-        if 'task_type' in self and value is not None:
+    def validate_extra_data(self, value):
+        if 'task_type' in self and self['task_type'] != '':
             validators.validate_task_type_with_meta_parameters(
                 field_name='meta_parameters',
                 task_type=self['task_type'],
-                meta_parameters=value,
+                meta_parameters=value['meta_parameters'],
                 task_type_meta_parameters=self.TASK_TYPE_META_PARAMETERS,
-            )
-        else:
-            validators.validate_correct_meta_parameters_class(
-                meta_parameters_classes=self.TASK_TYPE_META_PARAMETERS.values(),
-                value=value,
             )
 
     def validate_task_type(self, value):
-        if 'meta_parameters' in self and value is not None:
-            validators.validate_task_type_with_meta_parameters(
-                field_name='task_type',
-                task_type=value,
-                meta_parameters=self['meta_parameters'],
-                task_type_meta_parameters=self.TASK_TYPE_META_PARAMETERS,
-            )
-        else:
-            validators.validate_enum(
-                field_name='task_type',
-                value=value,
-                enum_class=TaskType
-            )
+        if value != '':
+            if 'extra_data' in self and 'meta_parameters' in self['extra_data']:
+                validators.validate_task_type_with_meta_parameters(
+                    field_name='task_type',
+                    task_type=value,
+                    meta_parameters=self['extra_data']['meta_parameters'],
+                    task_type_meta_parameters=self.TASK_TYPE_META_PARAMETERS,
+                )
+            else:
+                validators.validate_enum(
+                    field_name='task_type',
+                    value=value,
+                    enum_class=TaskType
+                )
 
 
 class TaskMessage(base.Message):
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -40,6 +40,45 @@ class BlenderScriptPackage(datastructures.ValidatingDict, datastructures.FrozenD
         'output_format': '',
     }
 
+    validate_resolution = functools.partial(
+        validators.validate_list_of_positive_integers,
+        field_name='resolution',
+        max_length=2,
+    )
+
+    validate_borders_x = functools.partial(
+        validators.validate_list_of_non_negative_floats,
+        field_name='borders_x',
+        max_length=2,
+    )
+
+    validate_borders_y = functools.partial(
+        validators.validate_list_of_non_negative_floats,
+        field_name='borders_y',
+        max_length=2,
+    )
+
+    validate_use_compositing = functools.partial(
+        validators.validate_boolean,
+        field_name='use_compositing'
+    )
+
+    validate_samples = functools.partial(
+        validators.validate_integer,
+        field_name='samples'
+    )
+
+    validate_frames = functools.partial(
+        validators.validate_list_of_positive_integers,
+        field_name='frames',
+    )
+
+    validate_output_format = functools.partial(
+        validators.validate_enum,
+        field_name='output_format',
+        enum_class=OUTPUT_FORMAT,
+    )
+
 
 class TaskType(datastructures.StringEnum):
     Blender = enum.auto()
@@ -80,6 +119,35 @@ class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
         field_name='subtask_id',
         max_length=128,
     )
+
+    def validate_meta_parameters(self, value):
+        if 'task_type' in self and value is not None:
+            validators.validate_task_type_with_meta_parameters(
+                field_name='meta_parameters',
+                task_type=self['task_type'],
+                meta_parameters=value,
+                task_type_meta_parameters=self.TASK_TYPE_META_PARAMETERS,
+            )
+        else:
+            validators.validate_correct_meta_parameters_class(
+                meta_parameters_classes=self.TASK_TYPE_META_PARAMETERS.values(),
+                value=value,
+            )
+
+    def validate_task_type(self, value):
+        if 'meta_parameters' in self and value is not None:
+            validators.validate_task_type_with_meta_parameters(
+                field_name='task_type',
+                task_type=value,
+                meta_parameters=self['meta_parameters'],
+                task_type_meta_parameters=self.TASK_TYPE_META_PARAMETERS,
+            )
+        else:
+            validators.validate_enum(
+                field_name='task_type',
+                value=value,
+                enum_class=TaskType
+            )
 
 
 class TaskMessage(base.Message):
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -18,6 +18,33 @@ from . import base
 TASK_MSG_BASE = 2000
 
 
+class OUTPUT_FORMAT(datastructures.StringEnum):
+    PNG = enum.auto()
+    EXR = enum.auto()
+    JPG = enum.auto()
+
+
+class BlenderScriptPackage(datastructures.ValidatingDict, datastructures.FrozenDict):
+    """
+    Represents data package for Blender rendering
+    Package is used to create python script
+    which is passed to Blender
+    """
+    ITEMS = {
+        'resolution': [],
+        'borders_x': [],
+        'borders_y': [],
+        'use_compositing': False,
+        'samples': '',
+        'frames': [],
+        'output_format': '',
+    }
+
+
+class TaskType(datastructures.StringEnum):
+    Blender = enum.auto()
+
+
 class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
     """Represents SUBTASK metadata."""
     ITEMS = {
@@ -28,11 +55,18 @@ class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
         # Task headers are received in MessageTasks.tasks.
         'deadline': 0,
         'src_code': '',
+        'script_template_hash': '',
+        'task_type': '',
+        'meta_parameters': '',
         'extra_data': {},  # safe because of copy in parent.__missing__()
         'short_description': '',
         'working_directory': '',
         'performance': 0,
         'docker_images': None,
+    }
+
+    TASK_TYPE_META_PARAMETERS = {
+        TaskType.Blender.name: BlenderScriptPackage,
     }
 
     validate_task_id = functools.partial(
--- a/golem_messages/validators.py
+++ b/golem_messages/validators.py
@@ -19,3 +19,107 @@ def validate_integer(field_name, value):
             field=field_name,
             value=value,
         )
+
+
+def validate_boolean(field_name, value):
+    if not isinstance(value, bool):
+        raise exceptions.FieldError(
+            f"Should be boolean",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_list(field_name, value, min_length, max_length):
+    if not (isinstance(value, list) and
+            (min_length <= len(value) <= max_length)):
+        raise exceptions.FieldError(
+            f"Should be a list with minimum {min_length} "
+            f"element and maximum {max_length} elements.",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_list_of_positive_integers(
+        field_name,
+        value,
+        min_length=1,
+        max_length=float('inf')
+):
+    validate_list(field_name, value, min_length, max_length)
+    for element in value:
+        if not (isinstance(element, int) and element > 0):
+            raise exceptions.FieldError(
+                f"Should be an integer greater than 0.0",
+                field=field_name,
+                value=value
+            )
+
+
+def validate_list_of_non_negative_floats(
+        field_name,
+        value,
+        min_length=1,
+        max_length=float('inf')
+):
+    validate_list(field_name, value, min_length, max_length)
+    for element in value:
+        if not (isinstance(element, float) and float(element) >= 0.0):
+            raise exceptions.FieldError(
+                f"Should be a float equal or greater than 0.0",
+                field=field_name,
+                value=value
+            )
+
+
+def validate_enum(field_name, value, enum_class):
+    try:
+        enum_class(value)
+    except ValueError:
+        raise exceptions.FieldError(
+            f" Invalid value for enum {enum_class.__name__}",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_task_type_with_meta_parameters(
+        field_name,
+        task_type,
+        meta_parameters,
+        task_type_meta_parameters,
+):
+    try:
+        task_class = task_type_meta_parameters[task_type]
+        task_class(meta_parameters)
+    except (exceptions.FieldError, ValueError, KeyError):
+        raise exceptions.FieldError(
+            f"Given task type: {task_type} is incompatible "
+            f"with meta parameters provided: {meta_parameters}.",
+            field=field_name,
+            value=(task_type, meta_parameters),
+        )
+
+
+def validate_correct_meta_parameters_class(meta_parameters_classes, value):
+    if value is None:
+        raise exceptions.FieldError(
+            'meta_parameters should be an instance one of the'
+            f'{meta_parameters_classes} classes',
+            field='meta_parameters',
+            value=value,
+        )
+    error_counter = 0
+    for script_package in meta_parameters_classes:
+        try:
+            script_package(value)
+        except KeyError:
+            error_counter += 1
+        if error_counter == len(meta_parameters_classes):
+            raise exceptions.FieldError(
+                'meta_parameters should be an instance one of '
+                f'{meta_parameters_classes} classes',
+                field='meta_parameters',
+                value=value,
+            )
--- a/tests/message/test_tasks.py
+++ b/tests/message/test_tasks.py
@@ -78,7 +78,7 @@ class ComputeTaskDefTestCase(unittest.TestCase):
         )
 
     def test_type(self):
-        ctd = message.ComputeTaskDef()
+        ctd = factories.tasks.ComputeTaskDefFactory()
         ctd['src_code'] = "custom code"
         msg = factories.tasks.TaskToComputeFactory(compute_task_def=ctd)
         s = msg.serialize()
@@ -96,9 +96,6 @@ class ComputeTaskDefTestCase(unittest.TestCase):
         self.assertTrue(extra_data['total_tasks'])
         self.assertTrue(extra_data['outfilebasename'])
         self.assertTrue(extra_data['scene_file'])
-        self.assertTrue(extra_data['script_src'])
-        self.assertTrue(extra_data['frames'])
-        self.assertTrue(extra_data['output_format'])
 
 
 class SubtaskResultsAcceptedTest(mixins.RegisteredMessageTestMixin,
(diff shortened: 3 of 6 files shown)


tests/test_datastructures.py
```diff
@@ -128,3 +128,50 @@ def test_set_fail(self):
...
+        ITEMS = {
+            'intvalue': 0,
+            'stringvalue': ''
+        }
+
+        DEPENDENT_ITEMS = [
+            ('intvalue', 'stringvalue'),
+        ]
+
+        def validate_intvalue_stringvalue(self, value):
```
dybi: Quite frankly, this validation tests two fields separately. Maybe you could make some validation, which would test interaction between these two fields?
I.e.:
```python
def validate_intvalue_stringvalue(self, value):
    if value['intvalue'] != int(value['stringvalue']):
        raise exceptions.FieldError()
```

tests/test_validators.py
```diff
@@ -48,3 +52,257 @@ def test_value_valid(self):
     def test_value_invalid_type(self):
         with self.assertRaises(exceptions.FieldError):
             self.validate(fake.word())
+
+
+class ValidatePositiveNumberValueTestCase(unittest.TestCase):
```
dybi: I don't mean to be picky, but I would prefer using `self.assertTrue` / `self.assertFalse`  in the tests - I perceive it more readable.

tests/test_validators.py
```diff
@@ -48,3 +52,257 @@ def test_value_valid(self):
...
+
+class ValidateTaskTypeWithMetaParametersTestCase(unittest.TestCase):
+    def setUp(self):
+        self.field_name = fake.word()
+        self.valid_pairs = [
+            (tasks.TaskType.Blender, tasks.BlenderScriptPackage),
+            (tasks.TaskType.LuxRender, tasks.LuxRenderScriptPackage),
+        ]
+        self.blender_script_package = BlenderScriptPackageFactory()
+        self.correct_value = {
```
dybi: What about `LuxRender`? It might be worth checking as well ;)

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
 
 
+class ValidatingDependentDictItems(ValidatingDict):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.validate_dependent_dict_items()
+
+    def validate_dependent_dict_items(self):
+        dependent_items_list = getattr(self, 'DEPENDENT_ITEMS', None)
```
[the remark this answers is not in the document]
rwrzesien: Why not just make `DEPENDENT_ITEMS` class attribute with empty list as default?

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
+        for items in dependent_items_list:  # pylint: disable=not-an-iterable
+            validator_function_dict = dict()
+            assert len(items) == 2
+            validator_function_name = "validate_" + "_".join(items)
```
rwrzesien: Use `f''` here if number of `items` is known and low.

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.validate_dependent_dict_items()
+
+    def validate_dependent_dict_items(self):
+        dependent_items_list = getattr(self, 'DEPENDENT_ITEMS', None)
+        for items in dependent_items_list:  # pylint: disable=not-an-iterable
+            validator_function_dict = dict()
+            assert len(items) == 2
```
rwrzesien: Maybe move `assert` one line above, it will be more readable.

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
             field=field_name,
             value=value,
         )
+
+
+def validate_positive_number_value(value, value_type):
```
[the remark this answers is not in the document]
kbeker: Change to `is_number_positive`

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+            (min_length <= len(value) <= max_length)):
+        raise exceptions.FieldError(
+            f"{field_name} should be a list with minimum {min_length}"
+            f" element and maximum {max_length} elements.",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_list_of_integers(
```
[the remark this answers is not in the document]
kbeker: Done

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+    for element in value:
+        if not isinstance(element, int) and element > 0:
+            raise exceptions.FieldError(
+                f"{field_name} should be a list with positive integers",
+                field=field_name,
+                value=value
+            )
+
+
+def validate_list_of_floats(
```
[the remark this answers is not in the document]
kbeker: Done

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+
+def validate_list_of_integers(
+        field_name,
+        value,
+        min_length=1,
+        max_length=float('inf')
+):
+    validate_list(field_name, value, min_length, max_length)
+    for element in value:
+        if not isinstance(element, int) and element > 0:
```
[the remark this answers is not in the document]
kbeker: Changed to -> `is_number_positive`

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+
+def validate_list_of_floats(
+        field_name,
+        value,
+        min_length=1,
+        max_length=float('inf')
+):
+    validate_list(field_name, value, min_length, max_length)
+    for element in value:
+        if not isinstance(element, float) and element >= 0:
```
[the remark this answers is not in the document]
kbeker: Done

golem_messages/message/tasks.py
```diff
@@ -17,7 +17,124 @@
 TASK_MSG_BASE = 2000
 
 
-class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
+class OUTPUT_FORMAT(datastructures.StringEnum):
+    PNG = enum.auto()
+    LXR = enum.auto()
```
cameel: Isn't this list incomplete? What about JPEG or EXR?

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+        )
+
+
+def validate_task_type_with_meta_parameters(field_name, value, valid_pairs):
+    for pair in valid_pairs:
+        if pair[0].name == value['task_type']:
+            try:
+                pair[1](value['meta_parameters'])
+            except KeyError:
+                raise exceptions.FieldError(
```
cameel: This `try ... except` is too broad. Do you expect to catch the `KeyError`:
- from `pair[1]` (when `pair` less than two items)?
- from `value['meta_parameters']` (when `value` does not have that key)?
- from inside the function stored in `pair[1]`?

Only the one from the three you want to cover should be inside the block. The others should be outside because otherwise you'll return a misleading error message.

tests/message/test_tasks.py
```diff
@@ -653,3 +651,197 @@ class TaskFailureTest(
...
+        self.assertIsInstance(blender_script_package, message.tasks.BlenderScriptPackage)
+        self.assertIsInstance(blender_script_package['resolution'], list)
+        self.assertIsInstance(blender_script_package['borders_x'], list)
+        self.assertIsInstance(blender_script_package['borders_y'], list)
+        self.assertIsInstance(blender_script_package['use_compositing'], bool)
+        self.assertIsInstance(blender_script_package['samples'], int)
+        self.assertIsInstance(blender_script_package['frames'], list)
+        self.assertIsInstance(blender_script_package['output_format'], message.tasks.OUTPUT_FORMAT)
+
+    def test_that_resolution_validation_will_raise_exception(self):
```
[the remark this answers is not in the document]
kbeker: Done

tests/message/test_tasks.py
```diff
@@ -653,3 +651,197 @@ class TaskFailureTest(
...
+        with self.assertRaises(exceptions.FieldError):
+            self.FACTORY(resolution=self.positive_int_tuple)
+
+        with self.assertRaises(exceptions.FieldError):
+            self.FACTORY(resolution=self.positive_float_list)
+
+    def test_that_resolution_validation_will_not_raise_exception(self):
```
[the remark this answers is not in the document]
kbeker: Done

tests/message/test_tasks.py
```diff
@@ -653,3 +651,197 @@ class TaskFailureTest(
     MSG_CLASS = message.tasks.TaskFailure
     FACTORY = factories.tasks.TaskFailureFactory
     TASK_ID_PROVIDER = 'task_to_compute'
+
+
+class BlenderScriptPackageTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
kbeker: Done

tests/message/test_tasks.py
```diff
@@ -653,3 +651,197 @@ class TaskFailureTest(
...
+
+        with self.assertRaises(exceptions.FieldError):
+            self.FACTORY(output_format=1)
+
+    def test_that_output_format_validation_will_not_raise_exception(self):
+        self.FACTORY(output_format=message.tasks.OUTPUT_FORMAT.PNG.name)
+        self.FACTORY(output_format=message.tasks.OUTPUT_FORMAT.LXR.name)
+
+
+class LuxRenderScriptPackageTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
kbeker: Done

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
             self, 'validate_{}'.format(key), None
         )
-
         if callable(validator):
             validator(value=value)  # noqa pylint:disable=not-callable
 
         super().__setitem__(key, value)
 
 
+class ValidatingDependentDictItems(ValidatingDict):
```
shadeofblue: @kbeker please add a docstring explaining the purpose of this class, etc

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.validate_dependent_dict_items()
+
+    def validate_dependent_dict_items(self):
+        dependent_items_list = getattr(self, 'DEPENDENT_ITEMS', [])
```
shadeofblue: @kbeker wouldn't it be better to verify if `DEPENDENT_ITEMS` is set in the `__init__` and initialize it there to an empty list if it's not? then you wouldn't need this check here and, as a bonus, you wouldn't need to escape pylint's complaint about `DEPENDENT_ITEMS` not being an iterable below...

golem_messages/validators.py
```diff
@@ -19,3 +19,103 @@ def validate_integer(field_name, value):
...
+    except ValueError:
+        raise exceptions.FieldError(
+            f"{field_name} has not valid value for"
+            f" {enum_class.__name__} enum class",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_task_type_with_meta_parameters(field_name, value, valid_pairs):
```
shadeofblue: @kbeker docstring please -> and generally, it would be good to have docstrings for each function in general ... they can be optional if the name of the method unambiguously and perspicuously describes the purpose... otherwise, it's good for everyone reading the code to have a clear explanation, especially in case of reusable stuff like validators...

tests/test_validators.py
kbeker: Done

tests/test_validators.py
kbeker: Added

golem_messages/datastructures.py
kbeker: Yes it work. I create this class to validate fields which are connected. Also validation works. There are tests for it.

kbeker: >Use f'' here if number of items is known and low. 

Done

golem_messages/message/tasks.py
kbeker: Actually it was. Now it is added and fixed LXR -> LXS

golem_messages/datastructures.py
shadeofblue: @kbeker but anyway, after reading the code more I, honestly, don't understand the real purpose of this class and I would lean towards not using it at all ... 

the only real usage in the code (apart from the very artificial usage in the test) is ensuring that `meta_parameters` are compatible with a particular `task_type`  ... which could, honestly, be achieved by just adding an explicit call to `validate_task_type_with_meta_parameters` validator to the `__init__` of the particular dict class (`ComputeTaskDef` in this case) that would explicitly use `VALID_TASK_TYPE_PAIRS` (which, btw, I'd name e.g. `TASK_TYPE_META_PARAMETERS` so that it's completely obvious what that refers to without the need to be aware of the way both `ValidatingDependentDictItems` and the validator itself work) ... and I'd make `TASK_TYPE_META_PARAMETERS` a dict, not a list of tuples since it's more natural to specify such mappings imvho...

golem_messages/datastructures.py
```diff
@@ -68,13 +68,35 @@ def __setitem__(self, key, value):
...
+        for items in dependent_items_list:  # pylint: disable=not-an-iterable
+            validator_function_dict = dict()
+            assert len(items) == 2
```
shadeofblue: and again, apart from the above - why would we use a list of tuples and then assert that each entry is 2 items long instead of just using a dict for a mapping - which would be both a lot more straightforward and readable and would eliminate the need to use such artificial asserts? ...

golem_messages/validators.py
shadeofblue: and even looking at the code here, I'm coming towards a conclusion that the usage of the `ValidatingDependentDictItems` is an unnecessary complication ... why use a generalized interface if we later strictly assume the field names? ... it just doesn't seem to make sense to me...

tests/test_validators.py
```diff
@@ -48,3 +52,257 @@ def test_value_valid(self):
...
+            value=value,
+            value_type=value_type,
+        )
+
+    def test_that_positive_int_value_returns_true(self):
+        true_value = self.validate(
+            value=self.positive_int_value,
+            value_type=int,
+        )
+        self.assertEqual(true_value, True)
```
shadeofblue: exactly, `self.assertTrue(true_value)` ... 

or just: `self.assertTrue(self.validate(....))` ...

golem_messages/datastructures.py
kbeker: >which could, honestly, be achieved by just adding an explicit call to validate_task_type_with_meta_parameters validator to the __init__ of the particular dict class

Actually this class is made because I didn't want to divide validation to different places. I saw that you made `ValidatingDict` to validate messages when they are initiating so I thought that the best way to validate `meta_parameters` and `task_type` would be in another class which will do the same thing but with multiple parameters. If you want I can change it, but individually I think that it should be in separate class like `ValidatingDependentDictItems` than in the message class

>VALID_TASK_TYPE_PAIRS (which, btw, I'd name e.g. TASK_TYPE_META_PARAMETERS

Names are of course to discuss, this was just a propose and it can be changed :) 

>and I'd make TASK_TYPE_META_PARAMETERS a dict, not a list of tuples since it's more natural to specify such mappings imvho...

Also I will change it

golem_messages/datastructures.py
kbeker: I have change it just to: 
```python
for items in getattr(self, 'DEPENDENT_ITEMS', []):
```
It doesn't give warning any more and works as it should work.

golem_messages/datastructures.py
kbeker: Yes, but you were talking about
```python
    TASK_TYPE_META_PARAMETERS = {
        TaskType.Blender.name: BlenderScriptPackage,
        TaskType.LuxRender.name: LuxRenderScriptPackage,
    }
```
which I have already changed from list to dict.

But, this assert is for 
```python
    DEPENDENT_ITEMS = [
        ('task_type', 'meta_parameters'),
    ]
```
to be sure that method `validate_dependent_dict_items` has `DEPENDENT_ITEMS` as a list of two item's tuples.

To be honest I don't really know how you want here create dict instead of tuple. Can you specify it more please?

golem_messages/datastructures.py
kbeker: This has been changed. Currently this class does not exist

golem_messages/datastructures.py
kbeker: Also this has been changed and does not exist in current version.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+
+def validate_task_type_with_meta_parameters(
+        field_name,
+        task_type,
+        meta_parameters,
+        task_type_class,
+        task_type_meta_parameters,
+):
+    try:
+        task_type_class(task_type)
```
dybi: this line should be removed from this validation - it is being checked in `validate_task_type` validator of `ComputeTaskDef`

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+        task_class = task_type_meta_parameters[task_type]
+        task_class(meta_parameters)
+    except (ValueError, KeyError):
+        raise exceptions.FieldError(
+            f"'task_type' has not valid value for"
```
dybi: This description is misleading - as I understand, the purpose of this validator is to check if `task_type` is compatible with `meta_parameters`. So the message should more like:
```python
f"Invalid meta parameters for task type: {task_type}"
```
or
```python
f"Given task type: {task_type} is incompatible with meta parameters provided: {meta_parameters}."
```

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+
+def validate_correct_meta_parameters_class(meta_parameters_classes, value):
+    if value is not None:
+        error_counter = 0
+        for script_package in meta_parameters_classes:
+            try:
+                script_package(value)
+            except KeyError:
+                error_counter += 1
+            if error_counter == len(meta_parameters_classes):
```
dybi: why? shouldn't it look more like:
```python
def validate_correct_meta_parameters_class(meta_parameters_classes, value):
    error_counter = 0
    for script_package in meta_parameters_classes:
        try:
            script_package(value)
        except KeyError:
            error_counter += 1
    if error_counter > 0:
        raise exceptions.FieldError(
            'meta_parameters should be an instance one of '
            f'{meta_parameters_classes} classes',
            field='meta_parameters',
            value=value,
        )
```
`None` check seems unnecessary.

tests/message/test_messages.py
```diff
@@ -345,3 +345,10 @@ def test_message_remove_task_container(self):
             )
         for i in range(test_cases):
             self.assertEqual(msg_l.remove_tasks[i].task_id, task_ids[i])
+
+    def test_message_task_to_compute_with_luxrender_script_package(self):
```
dybi: what about `blender_script_package`? It seems that there is no test for that.

golem_messages/validators.py
```diff
@@ -1,7 +1,7 @@
 from golem_messages import exceptions
 
 
-def validate_varchar(field_name, value, max_length):
+def validate_varchar(field_name, value, max_length=float('inf')):
```
rwrzesien: I think that instead of default `float('inf')`, you should check if `max_length` is not None, and then validate against it, otherwise no.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
             field=field_name,
             value=value,
         )
+
+
+def is_number_positive(value, value_type):
+    return isinstance(value, value_type) and value > 0
```
rwrzesien: If this will be called like ie. `is_number_positive('aaa', str)` it will raise exception. If that is ok, then ok, otherwise maybe validate if `value_type` is one of expected types?

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+    if not is_number_positive(value, int):
+        raise exceptions.FieldError(
+            "should be a positive integer field",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_not_negative_integer_value(field_name, value):
+    if not isinstance(value, int) or value < 0:
```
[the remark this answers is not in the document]
rwrzesien: Same here.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+        raise exceptions.FieldError(
+            f"{field_name} should be a list with minimum {min_length}"
+            f" element and maximum {max_length} elements.",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_list_of_positive_integers(
+        field_name,
```
rwrzesien: Please fix indentation.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+        if not is_number_positive(element, int):
+            raise exceptions.FieldError(
+                f"{field_name} should be a list with positive integers",
+                field=field_name,
+                value=value
+            )
+
+
+def validate_list_of_non_negative_floats(
+        field_name,
```
rwrzesien: Same here.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+        raise exceptions.FieldError(
+            f"{field_name} has not valid value for"
+            f" {enum_class.__name__} enum class",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_task_type_with_meta_parameters(
+        field_name,
```
rwrzesien: Same here.

tests/message/test_tasks.py
```diff
@@ -14,6 +14,7 @@
 from golem_messages import factories
 from golem_messages import message
 from golem_messages import shortcuts
+from golem_messages.factories.tasks import ComputeTaskDefFactory
```
rwrzesien: Factories are already imported.

tests/test_basic.py
```diff
@@ -16,6 +16,7 @@
 from golem_messages import datastructures
 
 from golem_messages.factories.helpers import clone_message
+from golem_messages.factories.tasks import ComputeTaskDefFactory
```
rwrzesien: Same here.

golem_messages/validators.py
```diff
@@ -19,3 +19,133 @@ def validate_integer(field_name, value):
...
+
+def validate_list_of_non_negative_floats(
+        field_name,
+        value,
+        min_length=1,
+        max_length=float('inf')
+):
+    validate_list(field_name, value, min_length, max_length)
+    for element in value:
+        if not is_number_positive(element, float) and float(element) != 0.0:
```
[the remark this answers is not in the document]
kbeker: Right :) Done

golem_messages/validators.py
kbeker: Done

golem_messages/validators.py
kbeker: Added assert to be sure that function can handle only integers and floats

golem_messages/validators.py
kbeker: Honestly I have did it in right way first, but when I run linter it gave me:
```
-      {
-          "type": "convention",
-          "module": "golem_messages.validators",
-          "obj": "",
-          "line": 68,
-          "column": 0,
-          "path": "golem_messages/validators.py",
-          "symbol": "bad-continuation",
-          "message": "Wrong hanging indentation before block (add 4 spaces).\n    field_name,\n    ^   |",
-          "message-id": "C0330"
-      },
```
That's why I left it with 8 spaces.

golem_messages/validators.py
kbeker: Same as above

golem_messages/validators.py
kbeker: Same as above

tests/test_basic.py
kbeker: Right, deleted. Done :)

tests/message/test_tasks.py
kbeker: Done

golem_messages/validators.py
kbeker: You're right. Deleted

tests/message/test_messages.py
kbeker: It is tested in every other messages because `blender_script_package` is a default package for `ComputeTaskDef`. But yea, why not. Never enough tests :) Done

golem_messages/validators.py
kbeker: It has to be `error_counter == len(meta_parameters_classes)` because I can't choose at first which script_package should be validate, so I need to iterate one by one and if none of these classes will be correct for script_packages which I got then it should raise exception

golem_messages/validators.py
kbeker: I thought a little bit about it and I have decided to delete `float('inf')` from here.

golem_messages/validators.py
dybi: right - at least one will always raise... :+1:

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
...
+                value=element
+            )
+
+
+def validate_enum(field_name, value, enum_class):
+    try:
+        enum_class(value)
+    except ValueError:
+        raise exceptions.FieldError(
+            f"{field_name} has not valid value for"
```
jiivan: `is not valid value for` ?

jiivan: Generally it's a little buttery butter :) *FieldError* already contains `field_name` and `value`. IMO sufficient message would be `Invalid value for enum {enum name}`.

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
...
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_list(field_name, value, min_length, max_length):
+    if not (isinstance(value, list) and
+            (min_length <= len(value) <= max_length)):
+        raise exceptions.FieldError(
+            f"{field_name} should be a list with minimum {min_length}"
```
jiivan: `field_name` duplicated

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
...
+            "should be a not negative integer field",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_boolean(field_name, value):
+    if not isinstance(value, bool):
+        raise exceptions.FieldError(
+            f"{field_name} should be boolean",
```
jiivan: `field_name` duplicated

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
             field=field_name,
             value=value,
         )
+
+
+def is_number_positive(value, value_type):
+    assert value_type in [int, float]
+    return isinstance(value, value_type) and value > 0
```
jiivan: Shouldn't this just throw *FieldError*?

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
...
+            except KeyError:
+                error_counter += 1
+            if error_counter == len(meta_parameters_classes):
+                raise exceptions.FieldError(
+                    'meta_parameters should be an instance one of '
+                    f'{meta_parameters_classes} classes',
+                    field='meta_parameters',
+                    value=value,
+                )
+    else:
```
jiivan: If you switch this condition to:
```python
if value is None:
    raise

error_counter....
```
then you'll simplify IMHO.

tests/message/test_messages.py
jiivan: IMO this should be moved to `tests/message/te_tasks.py`

tests/message/test_messages.py
```diff
@@ -345,3 +345,20 @@ def test_message_remove_task_container(self):
...
+        compute_task_def = factories.tasks.ComputeTaskDefFactory.with_luxrender_script_package()
+        task_to_compute = factories.tasks.TaskToComputeFactory(compute_task_def=compute_task_def)
+        serialized_message = shortcuts.dump(task_to_compute, None, None)
+        deserialzied_message = shortcuts.load(serialized_message, None, None)
+        self.assertEqual(task_to_compute, deserialzied_message)
+
+    def test_message_task_to_compute_with_blender_script_package(self):
```
jiivan: IMO this should be moved to `tests/message/te_tasks.py`

tests/message/test_tasks.py
```diff
@@ -694,3 +691,201 @@ class TaskFailureTest(
     MSG_CLASS = message.tasks.TaskFailure
     FACTORY = factories.tasks.TaskFailureFactory
     TASK_ID_PROVIDER = 'task_to_compute'
+
+#  pylint: disable=line-too-long
```
jiivan: I don't think it's justified here...

golem_messages/validators.py
```diff
@@ -19,3 +19,132 @@ def validate_integer(field_name, value):
...
+        raise exceptions.FieldError(
+            'meta_parameters should be an instance one of '
```
shadeofblue: should be an instance of one of the

golem_messages/validators.py
kbeker: Right, changed.

golem_messages/validators.py
kbeker: Done

golem_messages/validators.py
kbeker: Done

golem_messages/validators.py
kbeker: Done

golem_messages/validators.py
kbeker: Done

golem_messages/validators.py
kbeker: It shouldn't. Purpose of this function is just to give `true`/`false` answer. It is reused few times that's why I add it as separate function.

tests/message/test_tasks.py
kbeker: I added it here because I wanted to create long understandable for anyone tests function names. :)

tests/message/test_messages.py
kbeker: Oki, moved.

tests/test_validators.py
```diff
@@ -48,3 +54,301 @@ def test_value_valid(self):
...
+
+
+class ValidatePositiveNumberValueTestCase(unittest.TestCase):
+    def setUp(self):
+        self.positive_int_value = fake.random_int(min=1)
+        self.not_positive_int_value = fake.random_int(min=1) * (-1)
+        self.positive_float_value = fake.pyfloat(positive=True)
+        self.not_positive_float_value = fake.pyfloat(positive=True) * (-1)
+
+    def validate(self, value, value_type):  # pylint:disable=no-self-use
```
jiivan: 1. pylint:disable is unjustified here. (Could use `@classmethod`)
1. Only effect of this method is to rename `validators.is_number_positive` -> `self.validate`. IMO it should be removed

golem_messages/validators.py
```diff
@@ -19,3 +19,155 @@ def validate_integer(field_name, value):
...
+    if not (is_number_positive(value, int) or value == 0):
+        raise exceptions.FieldError(
+            "should be a positive integer field",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_not_negative_integer_value(field_name, value):
+    if not (is_number_positive(value, int) or value == 0):
```
jiivan: Duplicates `validate_positive_integer_value`

golem_messages/validators.py
```diff
@@ -19,3 +19,155 @@ def validate_integer(field_name, value):
...
+        raise exceptions.FieldError(
+            "should be a not negative integer field",
+            field=field_name,
+            value=value,
+        )
+
+
+def validate_not_negative_number(field_name, value):
```
jiivan: Duplicates `is_number_positive`.

golem_messages/validators.py
jiivan: Coming back to this. Please correct me if I'm wrong, but every time you use this function, you raise `FieldError` immediately when it returns `False`. That's why I insist it should just raise it, and be renamed to `verify_number_positive`.

Also every time except once, you check whether it's grater than *or equal* to zero. That's why this function should raise when `value < 0` and add additional check for *zero* in `validate_list_of_positive_integers`.

golem_messages/validators.py
```diff
@@ -19,3 +19,155 @@ def validate_integer(field_name, value):
...
+    for element in value:
+        if float(element) < 0.0:
+            raise exceptions.FieldError(
+                f"Should be a list of numbers equal or greater 0.0",
+                field=field_name,
+                value=element
+            )
+
+
+def validate_enum(field_name, value, enum_class):
```
jiivan: This somehow duplicates `Message.deserialize_slot()`:
https://github.com/golemfactory/golem-messages/blob/fb4afbc8e17265895adfb3059c67eb91c4062d70/golem_messages/message/base.py#L368-L378

Is it possible to combine those two approaches (reduce duplication)?

tests/message/test_tasks.py
jiivan: Could this be solved by splitting this test to subclasses?
For example:

1. `TestBlenderScriptPackageResolutionValidation(BlenderScriptPackageBase)`
   1. `test_not_positive_int_list()`
   1. `test_positive_int_list()`

tests/message/test_tasks.py
```diff
@@ -713,3 +710,208 @@ class TaskFailureTest(
...
+        self.assertIsInstance(blender_script_package['output_format'], message.tasks.OUTPUT_FORMAT)
+
+    def test_that_resolution_validation_will_raise_exception_when_resolution_is_not_positive_int_list(self):
+        with self.assertRaises(exceptions.FieldError):
+            self.FACTORY(resolution=self.positive_int_tuple)
+
+        with self.assertRaises(exceptions.FieldError):
+            self.FACTORY(resolution=self.positive_float_list)
+
+    def test_that_resolution_validation_will_not_raise_exception_when_is_positive_int_list_is_given(self):
```
jiivan: When is positive int list is given?

golem_messages/validators.py
kbeker: after removing all parts related to LuxRender problem solved. So, Done.

tests/test_validators.py
kbeker: Removed because of validators changes.

golem_messages/validators.py
kbeker: Deleted.

golem_messages/validators.py
kbeker: Removed because of validators changes.

tests/message/test_tasks.py
kbeker: @jiivan IMHO 80 characters for a line is a little bit not enough for a good test name. Can we change it for example to 120 chars per line?

tests/message/test_tasks.py
kbeker: Typo, fixed.

golem_messages/validators.py
kbeker: Hmm, it might be hard because `BlenderScriptPackage` doesn't inherit from `Message()` class where this method is implemented so I can't reuse `deserialzie_slot` in this dict.

shadeofblue:
R.I.P. ...

PR #274, opened 2018-10-23
Change price field in TaskToCompute

Krigpl:
Use value from `WantToComputeTask` and change its meaning to total amount to be paid, instead of rate per hour.

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -292,7 +292,6 @@ class TaskToCompute(ConcentEnabled, TaskMessage):
         'package_hash',  # the hash of the package (resources) zip file
         'size',  # the size of the resources zip file
         'concent_enabled',
-        'price',  # total subtask price computed as `price * subtask_timeout`
         'ethsig'
     ] + base.Message.__slots__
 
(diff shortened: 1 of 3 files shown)


shadeofblue:
@Krigpl did we decide anything with regards to this pull request? is there a complementary pull request for golem core?

Krigpl:
@shadeofblue  It's postponed until `TaskHeader` is included in `WantToComputeTask`

golem_messages/message/tasks.py
```diff
@@ -292,12 +291,16 @@ def provider_ethereum_public_key(self):
...
+    @property
+    def price(self):
+        return self.want_to_compute_task.price
+
     @base.verify_slot('want_to_compute_task', WantToComputeTask)
     def deserialize_slot(self, key, value):
         value = super().deserialize_slot(key, value)
         if key == 'compute_task_def':
             value = ComputeTaskDef(value)
-        if key in ('price', 'size'):
```
jiivan: Please move price validation to WTC along with price slot.

Krigpl:
I'd somewhat prefer a different name for this property though - since it doesn't represent rate per hour anymore.

jiivan:
> I'd somewhat prefer a different name for this property though - since it doesn't represent rate per hour anymore.

IMO WTC.price is more misleading. I propose `.rate_per_hour`

Krigpl:
> IMO WTC.price is more misleading. I propose .rate_per_hour

Following this convention TaskHeader.price should be changed as well. Most of Golem assumes hourly rate when using `price`

jiivan:
> Following this convention TaskHeader.price should be changed as well. Most of Golem assumes hourly rate when using `price`

It should.

shadeofblue:
actually, I think this pull request no longer makes sense

Krigpl:
@shadeofblue Why do you think that?

PR #281, opened 2018-11-21
Don't use custom Message CBOR encoder/decoder

Krigpl:
And switch to `cbor` library which uses C implementation, so it's more efficient.

current `cbor2`:
`$ python -m timeit -s "from golem_messages.factories.tasks import TaskToComputeFactory; from golem_messages.message.base import Message; ttc = TaskToComputeFactory()" "Message.deserialize(ttc.serialize(), decrypt_func=None)"
1000 loops, best of 3: 665 usec per loop`

this PR `cbor`:
`$ python -m timeit -s "from golem_messages.factories.tasks import TaskToComputeFactory; from golem_messages.message.base import Message; ttc = TaskToComputeFactory()" "Message.deserialize(ttc.serialize(), decrypt_func=None)"
1000 loops, best of 3: 266 usec per loop`


Krigpl:
I have no idea why the two tests are failing and it's hard for me to understand what is supposed to happen in these tests, any help?

jiivan:
> I have no idea why the two tests are failing and it's hard for me to understand what is supposed to happen in these tests, any help?

I suppose `test_fail_cannot_compute_task` checks wether `FieldError` is raised when `cannot_compute_task` is set to an instance of `TaskToCompute` (wrong class).
Similarly with `test_fail_task_failure`

tests/test_basic.py
```diff
@@ -490,9 +490,8 @@ def test_invalid_task_to_compute(self):
                 "There’s so much to learn when you’re slinging"
                 "paint and pencil"
             )
-            s = msg.serialize()
             with self.assertRaises(exceptions.FieldError):
-                message.Message.deserialize(s, decrypt_func=None)
+                msg.serialize()
```
jiivan: Will slot value be verified during deserialization phase? On the receivers side?

Krigpl: `deserialize_slot` works in a way that it will expect some message class, take the value and try to deserialize it into that class - I'd expect it will fail if the value is invalid, isn't that the case?

shadeofblue:
okay @Krigpl, I like those changes ... I'm just unsure why you removed all the tests that indeed verify if the validators work correctly - shouldn't there be new tests that check the equivalent conditions with the new mechanism?

PR #282, opened 2018-11-22
Remove result_type

Krigpl:
Not used in Golem


dybi:
Not used in `concent` either

Issue #284, opened 2018-12-05
Validate if `SubtaskResultsAccepted.payment_ts` <= Message.timestamp


shadeofblue:
add appropriate validation to `SubtaskResultsAccepted` and don't allow to deserialize a message that has an errorneous `payment_ts`

PR #286, opened 2018-12-05
Add validation of `payment_ts` in SubtaskResultsAccepted

dybi:
resolves: https://github.com/golemfactory/golem-messages/issues/284

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -197,6 +197,14 @@ class TaskMessage(base.Message):
         self.validate_ownership_chain(concent_public_key=concent_public_key)
         return True
 
+    @classmethod
+    def deserialize_with_header(cls, header, data, *args, **kwargs):  # noqa pylint: disable=arguments-differ
+        instance: SubtaskResultsAccepted = super().deserialize_with_header(
+            header, data, *args, **kwargs
+        )
+        instance.is_valid()
+        return instance
+
     def is_valid(self):  # noqa pylint:disable=no-self-use
         """
         checks whether the message is semantically valid with respect to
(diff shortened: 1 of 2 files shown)


golem_messages/message/tasks.py
```diff
@@ -483,6 +483,19 @@ class SubtaskResultsAccepted(TaskMessage):
         'report_computed_task': base.MessageSlot(ReportComputedTask),
     }
 
+    @classmethod
+    def deserialize_with_header(cls, header, data, *args, **kwargs):  # noqa pylint: disable=arguments-differ
+        instance: SubtaskResultsAccepted = super().deserialize_with_header(
+            header, data, *args, **kwargs
+        )
+        instance.is_valid()
```
shadeofblue: I actually think this should be part of `TaskMessage`'s `deserialize_with_header` (given the fact that `TaskMessage` contains the base implementation of `is_valid`)

golem_messages/message/tasks.py
```diff
@@ -483,6 +483,19 @@ class SubtaskResultsAccepted(TaskMessage):
...
+        return instance
+
+    def is_valid(self) -> bool:
```
shadeofblue: and, maybe indeed the base signature should be changed so that it's recognized `is_valid` can raise a `ValidationError`

dybi: @shadeofblue, afaik, there's no support for declaring throwing of exceptions in python. Or did you mean sth else?

golem_messages/message/tasks.py
dybi: done

golem_messages/message/tasks.py
shadeofblue: yeah, I mean the docstring syntax and actually, I see that it's already there ;)

```
        :raises: `exceptions.ValidationError`
```

golem_messages/message/tasks.py
```diff
@@ -197,6 +197,14 @@ def assert_role(role, expected, actual):
         self.validate_ownership_chain(concent_public_key=concent_public_key)
         return True
 
+    @classmethod
+    def deserialize_with_header(cls, header, data, *args, **kwargs):  # noqa pylint: disable=arguments-differ
+        instance: SubtaskResultsAccepted = super().deserialize_with_header(
```
shadeofblue: @dybi the type is _not_ guaranteed to be `SubtaskResultsAccepted` here ...

dybi: good catch, I simply moved it from `SubtaskResultsAccepted` without changing the type... my bad ;)

PR #293, opened 2018-12-17
Change allow_none value to False in ForcePayment message

kbeker:


cameel:
@kbeker When making pull requests like this, please first make an issue that explains the bug. I can't review it because I have no idea why you're changing this.

Issue #247, opened 2018-08-17
Update ethereum address proof to use a regular `ethsig` in TaskToCompute


shadeofblue:
Update ethereum address proof to use a regular `ethsig` slot on `TaskToCompute` and use the existing API to generate the signature by signing the included `WantToComputeTask` message (stored in `want_to_compute_task` field of `TaskToCompute`) instead of the whole `TaskToCompute`

depends on:
* [x] https://github.com/golemfactory/golem-messages/issues/241
* [x] https://github.com/golemfactory/golem-messages/issues/244

incidentally, it's a possible solution for #295

PR #298, opened 2018-12-20
Feature update ethereum address proof to use regular ethsig in task to compute

rwrzesien:
Resolves https://github.com/golemfactory/golem-messages/issues/247

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -1,6 +1,5 @@
 import enum
 import functools
-import struct
 import typing
 
 from eth_utils import to_checksum_address
@@ -263,8 +262,6 @@ class WantToComputeTask(ConcentEnabled, base.Message):
 @library.register(TASK_MSG_BASE + 2)
 class TaskToCompute(ConcentEnabled, TaskMessage):
     EXPECTED_OWNERS = (TaskMessage.OWNER_CHOICES.requestor, )
-    ETHSIG_FORMAT = '66p'
-    ETHSIG_LENGTH = struct.calcsize(ETHSIG_FORMAT)
 
     MSG_SLOTS = {
         'want_to_compute_task': base.MessageSlotDefinition(WantToComputeTask),
@@ -281,8 +278,7 @@ class TaskToCompute(ConcentEnabled, TaskMessage):
         'size',  # the size of the resources zip file
         'concent_enabled',
         'price',  # total subtask price computed as `price * subtask_timeout`
-
-        '_ethsig'  # must be last
+        'ethsig'
     ] + base.Message.__slots__
 
     @property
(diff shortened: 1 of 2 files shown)


golem_messages/message/tasks.py
```diff
@@ -330,33 +329,20 @@ def subtask_id(self):
...
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        if not hasattr(self, '_ethsig'):
-            self._ethsig = None
-
     def serialize(self, *args, **kwargs):  # pylint: disable=arguments-differ
         serialized = super().serialize(*args, **kwargs)
         sig_length = self.ETHSIG_LENGTH - 1
-        if self._ethsig and len(self._ethsig) != sig_length:
+        if self.ethsig and len(self.ethsig) != sig_length:
```
shadeofblue: @rwrzesien I don't think this is needed here -> this field is now serialized normally, alongside the other ones, so the length check isn't needed - if the signature is wrong - it just won't pass `verify_ethsig()` ...

golem_messages/message/tasks.py
```diff
@@ -388,7 +375,10 @@ def generate_ethsig(
                 value=self.requestor_ethereum_public_key,
             )
 
-        self._ethsig = self._get_signature(private_key, msg_hash)
+        if msg_hash is None and self.want_to_compute_task is not None:
```
shadeofblue: why? .... I believe:

`self.ethsig = self.want_to_compute_task._get_signature(private_key, msg_hash)` should do the trick...

rwrzesien: The question is if there should be still possibility to create `ethsig` with custom `msg_hash` passed to this function. As this possibility was there before, I have assumed it should still be there for optimization purposes, as in the method comment `to skip generation of the message hash while signing`. Should I remove it then?

rwrzesien: Not exactly, if `msg_hash` will be `None,` then `_get_signature` will try to create it from `self.get_short_hash()`, when `self` is `TaskToCompute` in this context.

golem_messages/message/tasks.py
rwrzesien: :+1: done

golem_messages/message/tasks.py
shadeofblue: well, not really, because `_get_signature()` will be called from `WantToComputeTask` (in `want_to_compute_task`) and not from `self.` ...

golem_messages/message/tasks.py
```diff
@@ -399,12 +389,17 @@ def verify_ethsig(
...
 
         :param msg_hash: maybe optionally provided to skip generation
-                         of the message hash during the verification
+                         of the message hash during the verification.
+                         If not provided, the hash is generated from nested
+                         WantToComputeTask.
         :return: `True` if the signature is correct.
         :raises: `exceptions.InvalidSignature` if the signature is corrupted
         """
+        if msg_hash is None and self.want_to_compute_task is not None:
```
shadeofblue: again, I believe ```
self.want_to_compute_task._verify_signature(
self.ethsig,
decode_hex(self.requestor_ethereum_public_key)`,
msg_hash)
```

should be enough + more readable

Comment (rwrzesien) [reply golem_messages/message/tasks.py]: :+1: done

File: golem_messages/message/tasks.py (hunk 2 above)
Comment (rwrzesien) [reply golem_messages/message/tasks.py]: :+1: done

File: golem_messages/message/tasks.py [hunk 4]
```diff
@@ -330,33 +329,11 @@ def subtask_id(self):
...
         return self
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        if not hasattr(self, '_ethsig'):
-            self._ethsig = None
-
-    def serialize(self, *args, **kwargs):  # pylint: disable=arguments-differ
-        serialized = super().serialize(*args, **kwargs)
-        sig_length = self.ETHSIG_LENGTH - 1
```
jiivan: `ETHSIG_LENGTH` & `ETHSIG_FORMAT` are not needed anymore (remove from class)

rwrzesien: :+1: done

Issue #300, opened 2018-12-21
Settlement algorithm should reject acceptances with large difference between timestamp and payment_ts


Jakub89:
This issue was created by @cameel in concent project. But we decide to implement it in here.
Original issue:
https://github.com/golemfactory/concent/issues/1036

PR #301, opened 2018-12-21
Add payment_ts validation in SubtaskResultAccepted message

Jakub89:
Resolves: #300

### DIFF
--- a/golem_messages/settings.py
+++ b/golem_messages/settings.py
@@ -4,7 +4,7 @@ import time
 
 MSG_TTL = datetime.timedelta(minutes=10)
 FUTURE_TIME_TOLERANCE = datetime.timedelta(minutes=5)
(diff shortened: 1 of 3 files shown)


golem_messages/settings.py
```diff
@@ -4,6 +4,7 @@
 
 MSG_TTL = datetime.timedelta(minutes=10)
 FUTURE_TIME_TOLERANCE = datetime.timedelta(minutes=5)
+PAYMENT_TIMESTAMP_TOLERANCE = datetime.timedelta(minutes=15).total_seconds()
```
jiivan: Since all tolerances are `timedelta`s I think one should also be a `timedelta`

Jakub89: Done

Issue #302, opened 2019-01-02
Actually include TaskHeader in WantToComputeTask


shadeofblue:
* Make sure `WantToComputeTask` includes the `TaskHeader` record signed by the requestor as one of its slots.
* before merging, please make sure the Golem Core counterpart: https://github.com/golemfactory/golem/issues/3696 is implemented and ready for merge

PR #303, opened 2019-01-03
Feature include task header in want to compute task

dybi:
resolves: https://github.com/golemfactory/golem-messages/issues/302
golem counterpart: https://github.com/golemfactory/golem/pull/3718


golem_messages/message/base.py
```diff
@@ -1,723 +1,728 @@
...
-        WrongEncryption = 'wrong_encryption'
-        ResourceHandshakeFailure = 'resource_handshake'
-        KeyNotDifficult = 'key_not_difficult'
-        Bootstrap = 'bootstrap'
-
-
-@library.register(3)
-class ChallengeSolution(Message):
-    __slots__ = ['solution'] + Message.__slots__
+import calendar
```
Krigpl: image

dybi: CRLF -> LF and I didn't notice... :(

tests/message/test_tasks.py
```diff
@@ -283,6 +285,26 @@ def test_no_compute_task_def(self):
         # Should not raise
         factories.tasks.TaskToComputeFactory(compute_task_def=None)
 
+    def test_validate_ownership_chain(self):
```
shadeofblue: @dybi I think a negative test would also be advised

golem_messages/message/base.py
Krigpl: If you're changing back to the correct line endings please do it in separate PR as it's awkward to see what has been changed in this file

golem_messages/message/base.py
```diff
@@ -261,6 +262,8 @@ def serialize_slot(self, key, value):
             return value.value
         if key in self.MSG_SLOTS:
             return self.serialize_message(key, value)
+        if isinstance(value, TaskHeader):
```
jiivan: in `deserialize_slot()` you're checking againts a key (`if key == 'task_header' and value is not None`) IMO the same should be done here.

golem_messages/message/tasks.py
```diff
@@ -249,7 +249,8 @@ class WantToComputeTask(ConcentEnabled, base.Message):
                             # with regards to the contents of this field.
 
         'provider_public_key',  # key used for msg signing and encryption
-        'provider_ethereum_public_key',  # used for transactions on blockchain
+        'provider_ethereum_public_key',  # used for transactions on blockchain,
```
jiivan:
```suggestion
        'provider_ethereum_public_key',  # used for transactions on blockchain
```

tests/message/test_messages.py
```diff
@@ -60,6 +61,7 @@ def test_message_want_to_compute_task(self):
         concent_enabled = random.random() > 0.5
         provider_public_key = provider_ethereum_public_key = (
             encode_hex(cryptography.ECCx(None).raw_pubkey))
+        task_header = TaskHeaderFactory()
         msg = message.WantToComputeTask(
```
jiivan: This whole method should be removed, and `tests/messages/test_tasks.py::WantToComputeTaskTest` should use `SerializationMixin`. But it's a little bit on the edge of this PR (not mandatory)

tests/message/test_tasks.py
```diff
@@ -197,6 +198,7 @@ def test_task_to_compute_basic(self):
         serialized = shortcuts.dump(ttc, None, None)
```
jiivan: This whole method should be remove. It's already covered by `SerializationMixin`

tests/message/test_tasks.py
```diff
@@ -283,6 +285,46 @@ def test_no_compute_task_def(self):
...
+            requestor_public_key=encode_hex(
+                requestor_keys.raw_pubkey,
+            ),
+            want_to_compute_task=wtc,
+        )
+        ttc.sign_message(requestor_keys.raw_privkey)  # noqa pylint: disable=no-value-for-parameter
+
+        ttc.validate_ownership_chain()
+
+    def test_validate_ownership_chain_rasie_when_invalid(self):
```
jiivan:
```suggestion
    def test_validate_ownership_chain_raises_when_invalid(self):
```

tests/message/test_messages.py
dybi: will do ;)

golem_messages/message/base.py
dybi: checking if value is a `TaskHeader` instance is safer -> we know that we can call `value.to_dict()`; having just the key name and not being `None` doesn't give us the same level of certainty...

jiivan: This way you will silently serialize message with wrong field values, i.e. a `TaskHeader` in a `provider_public_key` slot.
Also I just noticed it's implemented in `Base`. IMO it should be implemented as a mixin in `golem_messages.datastructures.tasks`. Similar to `golem_messages.datastructures.p2p.NodeSlotMixin`.

dybi: what about:
`if key == 'task_header' and isinstance(value, TaskHeader)` ? ;)

I will move it to 'WantToComputeTask'

golem_messages/datastructures/tasks.py
```diff
@@ -117,10 +117,10 @@ def sign(self, private_key: bytes) -> None:
             msghash=self.get_hash(),
         )
 
-    def verify(self, public_key: bytes) -> None:
+    def verify(self, public_key: bytes) -> bool:
```
jiivan: @shadeofblue this suggests that when verification fails it will return False. :(

golem_messages/factories/tasks.py
```diff
@@ -99,7 +159,18 @@ def task_id(
         if ttc.compute_task_def is None:
             return
 
-        ttc.compute_task_def['task_id'] = extracted or helpers.fake_golem_uuid(  # noqa pylint: disable=unsupported-assignment-operation
+        _task_id = None
+        with suppress(AttributeError):
```
jiivan:
```python
>>> import this
[...]
"There should be one-- and preferably only one --obvious way to do it."
```
I'm not convinced to this suppress thing :). It doesn't add anything to context.

golem_messages/factories/tasks.py
```diff
@@ -99,7 +159,18 @@ def task_id(
...
+            if ttc.requestor_id == \
+                    ttc.want_to_compute_task.task_header.task_owner.key:
+                _task_id = ttc.want_to_compute_task.task_header.task_id
+        # except AttributeError:
+        #     pass
```
jiivan: If you left this comments it only confirms that *suppress* is not bvious way of silencing exceptions.

golem_messages/message/base.py
```diff
@@ -300,6 +301,8 @@ def deserialize_slot(self, key, value):
                 ) from e
         if key in self.MSG_SLOTS:
             return self.deserialize_message(key, value)
+        if key == 'task_header' and value is not None:
+            return TaskHeader(**value)
```
jiivan: This is WTCT specific

golem_messages/factories/tasks.py
dybi: hm... I would say is more succinct and better expresses intention of doing nothing (ignoring) in case particular exceptions shows up

golem_messages/factories/tasks.py
dybi: no, it's just me being inattentive... sorry, I will remove it

PR #304, opened 2019-01-07
Fixed Node serialization in Peer

zakaprov:
Fixes: [#3705](https://github.com/golemfactory/golem/issues/3705)

When serializing Peer instances, the inner Node was being converted to
a dict and put back under the same key. Due to Peer's custom
__setitem__ implementation, the dict was converted back into a Node object upon insertion.
This caused later CBOR serialization to fail.

### DIFF
--- a/tests/datastructures/test_p2p.py
+++ b/tests/datastructures/test_p2p.py
@@ -4,6 +4,7 @@ import faker
 
 from golem_messages import exceptions
 from golem_messages.factories.datastructures import p2p as dt_p2p_factory
+from golem_messages import message
 
 
 fake = faker.Faker()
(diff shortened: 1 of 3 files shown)


tests/message/test_messages.py
```diff
@@ -362,3 +362,11 @@ def test_message_remove_task_container(self):
             )
         for i in range(test_cases):
             self.assertEqual(msg_l.remove_tasks[i].task_id, task_ids[i])
+
+    def test_peers_serialization(self):
```
jiivan: Please place it in `tests/message/test_p2p.py`

tests/message/test_messages.py
```diff
@@ -362,3 +362,11 @@ def test_message_remove_task_container(self):
...
+        peers = [dt_p2p_factory.Peer(), ]
```
etam: It's not a tuple. To make one element list, you don't need the comma.

tests/datastructures/test_p2p.py
```diff
@@ -91,3 +92,11 @@ def test_address_ipv6(self):
     def test_address_none(self):
         with self.assertRaises(exceptions.FieldError):
             self.peer['address'] = None
+
+    def test_peers_serialization(self):
+        peers = [dt_p2p_factory.Peer()]
```
etam:
```suggestion
        peers = [self.peer]
```

PR #305, opened 2019-01-08
Fixed WantToStartTaskSession serialization

zakaprov:
Updated order of inheritance to trigger node-specific serialization
before base one.

### DIFF
--- a/tests/message/test_p2p.py
+++ b/tests/message/test_p2p.py
@@ -1,6 +1,8 @@
 import unittest
+
 from golem_messages.factories import p2p as p2p_factories
 from golem_messages.message import p2p
+
 from tests.message import mixins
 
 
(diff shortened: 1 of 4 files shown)


tests/message/test_p2p.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+from golem_messages.factories import p2p
+from tests.message import mixins
+
+
+class WantToStartSessionTest(mixins.SerializationMixin, unittest.TestCase):
```
jiivan: As you're on it maybe use also RegisteredMessageTestMixin

tests/message/test_p2p.py
```diff
@@ -0,0 +1,11 @@
...
+    FACTORY = p2p.WantToStartTaskSessionFactory
+
+
+class SetTaskSessionTest(mixins.SerializationMixin, unittest.TestCase):
```
jiivan: the same

PR #306, opened 2019-01-09
Feature include task header in want to compute task a

shadeofblue:


golem_messages/datastructures/tasks.py
```diff
@@ -120,7 +120,7 @@ def sign_task(self, private_key: bytes) -> None:
     def verify(self, public_key: bytes) -> None:
```
dybi: please change the return type:
```suggestion
    def verify(self, public_key: bytes) -> None:
```
```suggestion
    def verify(self, public_key: bytes) -> bool:
```

golem_messages/factories/tasks.py
```diff
@@ -101,7 +163,18 @@ def task_id(
         if ttc.compute_task_def is None:
             return
 
-        ttc.compute_task_def['task_id'] = extracted or helpers.fake_golem_uuid(  # noqa pylint: disable=unsupported-assignment-operation
+        _task_id = None
+        try:
```
dybi: maybe: 
``` python
        with suppress(AttributeError):
            if ttc.requestor_id == \
                    ttc.want_to_compute_task.task_header.task_owner.key:
                _task_id = ttc.want_to_compute_task.task_header.task_id
```
instead? it's simpler/shorter

golem_messages/datastructures/tasks.py
jiivan: I disagree. The only interesting effect of this function is whether it raises or not.

golem_messages/datastructures/tasks.py
```diff
@@ -120,7 +120,7 @@ def sign_task(self, private_key: bytes) -> None:
...
         if self.signature is None:
             raise exceptions.InvalidSignature("No signature")
-        cryptography.ecdsa_verify(
+        return cryptography.ecdsa_verify(
```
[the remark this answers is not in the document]
shadeofblue: true, but on the other hand, it's natural to expect that value to be True in case the verification succeeds ... plus, I did that mostly for consistency with other `sign()` functions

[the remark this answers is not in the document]
jiivan: Diseases are also natural :) I would expect it to behave as declared in return type.

PR #308, opened 2019-01-11
Remove GetResource and ResourceList message

badb:
Information about resources should be included in ComputeTaskDef. Thanks to that we can:
1) Simplify the protocol
2) Add support for sending different resources for different subtasks, which will be necessary for future use cases.

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -32,6 +32,7 @@ class ComputeTaskDef(datastructures.ValidatingDict, datastructures.FrozenDict):
         'short_description': '',
         'performance': 0,
         'docker_images': None,
+        'resources': [],
     }
 
     validate_task_id = functools.partial(
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -443,15 +443,6 @@ class ReportComputedTask(TaskMessage):
     ] + base.Message.__slots__
 
 
-@library.register(TASK_MSG_BASE + 8)
-class GetResource(base.Message):
-    """Request a resource for a given task"""
-    __slots__ = [
-        'task_id',
-        'resource_header'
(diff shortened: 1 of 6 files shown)


golem_messages/message/tasks.py
```diff
@@ -293,6 +294,7 @@ class TaskToCompute(ConcentEnabled, TaskMessage):
         'size',  # the size of the resources zip file
```
Krigpl: You probably wanna remove this and the line above?

tests/message/test_tasks.py
```diff
@@ -109,6 +109,10 @@ def test_extra_data(self):
         self.assertTrue(extra_data['frames'])
         self.assertTrue(extra_data['output_format'])
 
+    def test_has_resources(self):
+        ctd = message.ComputeTaskDef()
```
jiivan:
```suggestion
        ctd = message.tasks.ComputeTaskDef()
```

jiivan: IMHO This method only tests whether `FrozenDict` works as expected. I think it's not necessary.

tests/message/test_tasks.py
```diff
@@ -348,6 +352,10 @@ def test_size_notint(self):
         with self.assertRaises(exceptions.FieldError):
             helpers.dump_and_load(ttc)
 
+    def test_resources_options(self):
+        ttc = factories.tasks.TaskToComputeFactory()
+        self.assertTrue(hasattr(ttc, 'resources_options'))
```
jiivan: IMHO this method only tests whether `messages.base.Message.load_slots()` works. It initializes all missing slots to `None`. I think it's not necessary to test this in this context.

golem_messages/message/tasks.py
Krigpl: Why not ?

PR #317, opened 2019-01-30
make `TaskHeader`'s string representation more useful

shadeofblue:


zakaprov:
Please add more tests.

Issue #320, opened 2019-02-14
Make tasks.WaitingForResults a task message


jiivan:
Add `task_to_compute` slot so `WaitingForResults` message so that it can be statelessly verified on provider's side.

In golem `TaskSession._react_to_waiting_for_results()` add:
```python
        if self.concent_service.available:
              concent_key = self.concent_service.variant['pubkey']
          else:
              concent_key = None
          msg.verify_owners(
              requestor_public_key=self.key_id,
              provider_public_ket=self.task_server.keys_auth.ecc.raw_pubkey,
              concent_public_key=concent_key,
          )
```

Alse remove setting `session.task_id` in `TaskServer.new_session_prepare()`

PR #321, opened 2019-02-15
WaitingForResults becomes a task message

jiivan:
fixes: #320 
complementary PR: https://github.com/golemfactory/golem/pull/3896

### DIFF
--- a/golem_messages/message/tasks.py
+++ b/golem_messages/message/tasks.py
@@ -561,8 +561,16 @@ class StartSessionResponse(base.Message):
 
 
 @library.register(TASK_MSG_BASE + 25)
-class WaitingForResults(base.Message):
-    __slots__ = base.Message.__slots__
+class WaitingForResults(TaskMessage):
+    TASK_ID_PROVIDERS = ('task_to_compute', )
+    EXPECTED_OWNERS = (TaskMessage.OWNER_CHOICES.requestor, )
+    MSG_SLOTS = {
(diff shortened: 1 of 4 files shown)


golem_messages/message/tasks.py
```diff
@@ -561,8 +561,16 @@ def __init__(self, conn_id=None, **kwargs):
 
 
 @library.register(TASK_MSG_BASE + 25)
-class WaitingForResults(base.Message):
-    __slots__ = base.Message.__slots__
+class WaitingForResults(TaskMessage):
+    TASK_ID_PROVIDERS = ('task_to_compute', )
+    EXPECTED_OWNERS = (TaskMessage.OWNER_CHOICES.requestor, )
+    MSG_SLOTS = {
+        'task_to_compute': base.MessageSlotDefinition(TaskToCompute),
```
Krigpl: Isn't `WaitingForResults` a reply to `WantToComputeTask`? Wouldn't it be clearer to include wtct instead?

jiivan: Then you won't get information for which task results requestor is waiting.

Krigpl: Then maybe include `ReportComputedTask` ?

jiivan: :) @shadeofblue edited #320 this morning and changed rtc to ttc. RTC won't always be available. WFR can happen anytime since subtask assignment.

PR #325, opened 2019-02-18
V2 to master

shadeofblue:


golem_messages/message/base.py
```diff
@@ -169,6 +169,10 @@ def __repr__(self):
             slots=slots_,
         )
 
+    @classmethod
```
jiivan: It could be a property. Then there would be no need to disable linters.

PR #336, opened 2019-04-15
Explicit provider ETH addr instead of derived from ETH pub key

tworec:
Provider ETH address used explicitly instead of a Provider's public key in WantToComputeTask

Resolves (partly) https://github.com/golemfactory/golem-unlimited/issues/217
complemetary to golemfactory/golem#4170


golem_messages/factories/tasks.py
```diff
@@ -23,9 +24,10 @@ class Meta:
     node_name = factory.Faker('name')
     provider_public_key = factory.LazyFunction(
         lambda: encode_key_id(cryptography.ECCx(None).raw_pubkey))
-    provider_ethereum_public_key = factory.SelfAttribute(
-        'provider_public_key'
-    )
+    provider_ethereum_address = factory.LazyFunction(
```
shadeofblue: @tworec please make sure it uses `provider_public_key` as the basis of the address and not a completely different key

golem_messages/message/tasks.py
```diff
@@ -234,31 +234,52 @@ def __init__(self, *args, **kwargs):
...
-                            # `golem-messages` should be intentionally agnostic
-                            # with regards to the contents of this field.
-
-        'provider_public_key',  # key used for msg signing and encryption
-        'provider_ethereum_public_key',  # used for transactions on blockchain
-        'task_header',  # TaskHeader, should be signed by requestor
+        'num_subtasks',
+        'concent_enabled',
+        'extra_data',
+        # TODO: `provider` prefix is redundant; all above fields refers Provider
```
shadeofblue: "explicit is better than implicit"

```suggestion
```

tworec: agreed, then lets consequently name all others with prefix `provider_`
this is even more valid in case of TTC, where some fields are from requestor and some from provider afaik

golem_messages/factories/tasks.py
tworec: this factory was wrong: `provider_public_key` has nothing to do with `provider_ethereum_public_key`, therefore  `provider_ethereum_address` shouldn't base on it also

golem_messages/message/tasks.py
```diff
@@ -234,31 +234,52 @@ def __init__(self, *args, **kwargs):
...
+        concent_enabled:    Provider's Concent status
+        extra_data:         additional required information about the Provider's
+                            environment. `golem-messages` should be
+                            intentionally agnostic with regards to the contents
+                            of this field.
+        provider_public_key:
+                            for signing and encryption
+        provider_ethereum_address:
+                            for transactions on ETH blockchain
+        task_header:        Demand; signed by a Requestor
```
shadeofblue: `TaskHeader` is a class name... so unless you're also planning to update the class name, please leave the "type hint" intact here... 

```suggestion
        task_header:        TaskHeader; a Demand signed by a Requestor
```

golem_messages/factories/tasks.py
```diff
@@ -23,9 +24,10 @@ class Meta:
...
+        lambda: to_checksum_address(
```
shadeofblue: @tworec you can now move on to use `pubkey_to_address` from `utils`

tests/datastructures/test_tasks.py
```diff
@@ -39,15 +39,15 @@ def test_validate_illegal_deadline(self):
         self.th_dict_repr['deadline'] = datetime.datetime.now()
         with self.assertRaisesRegex(
             exceptions.FieldError,
-            r"^Should be an integer \[deadline:datetime\.datetime\(.+\)\]$"
+            r"^Should be a integer \[deadline:datetime\.datetime\(.+\)\]$"
```
shadeofblue: an

tests/datastructures/test_tasks.py
```diff
@@ -39,15 +39,15 @@ def test_validate_illegal_deadline(self):
...
         ):
             dt_tasks.TaskHeader(**self.th_dict_repr)
 
     def test_validate_illegal_timeout(self):
         self.th_dict_repr['subtask_timeout'] = "abc"
         with self.assertRaisesRegex(
             exceptions.FieldError,
-            r"Should be an integer \[subtask_timeout:'abc'\]"
+            r"Should be a integer \[subtask_timeout:'abc'\]"
```
shadeofblue: an

tests/datastructures/test_tasks.py
```diff
@@ -87,7 +87,7 @@ def test_validate_subtasks_count_invalid_type(self):
         self.th_dict_repr['subtasks_count'] = None
         with self.assertRaisesRegex(
             exceptions.FieldError,
-            r"^Should be an integer \[subtasks_count:None\]$",
+            r"^Should be a integer \[subtasks_count:None\]$",
```
shadeofblue: an

tests/message/test_tasks.py
```diff
@@ -46,28 +46,70 @@ def test_extra_data(self):
...
+    def test_num_subtasks_more_than_1(self):
+        wtct = message.tasks.WantToComputeTask(num_subtasks=17)
+        self.assertEqual(wtct.num_subtasks, 17)
+
+    def test_num_subtasks_non_int_raises(self):
+        wtct = message.tasks.WantToComputeTask(num_subtasks=3.14)
+        serialized = shortcuts.dump(wtct, None, None)
+        with self.assertRaisesRegex(
+            exceptions.FieldError,
+            r"Should be a integer \[num_subtasks:3.14\]"
```
shadeofblue: an

golem_messages/factories/tasks.py
```diff
@@ -1,18 +1,19 @@
 # pylint: disable=too-few-public-methods,unnecessary-lambda
 import calendar
-from contextlib import suppress
 import datetime
 import time
 import typing
+from contextlib import suppress
```
jiivan: Why? Alphabetically it should be placed where it was before.

tworec: pycharm formatter sorted this way

golem_messages/factories/tasks.py
shadeofblue: @tworec I'll insist here ...

tests/message/test_tasks.py
```diff
@@ -48,17 +49,9 @@ def test_extra_data(self):
 
     def test_provider_ethereum_address_checksum(self):
         wtct = self.FACTORY()
-        self.assertTrue(wtct.provider_ethereum_public_key)
+        self.assertTrue(wtct.provider_ethereum_address)
         self.assertTrue(is_checksum_address(wtct.provider_ethereum_address))
 
-    def test_ethereum_address_provider(self):
```
shadeofblue: and pls, bring back this test here...

golem_messages/factories/tasks.py
tworec: ok

golem_messages/utils.py
```diff
@@ -2,17 +2,18 @@
...
+from eth_utils import to_checksum_address, keccak, hexstr_if_str
 
 from golem_messages import message
 
 
 def pubkey_to_address(pubkey: str) -> str:
     """
-    convert a hex-encoded pubkey into a checksummed address
+    convert pubkey into a checksummed address
+    pubkey might be hex str or bytes or int
```
Krigpl: No, we don't do that here. `pubkey` is a string, is typed as such and leave it this way.

tworec: why is that? can you provide some rationale?

Krigpl: throw-anything-as-an-argument-and-hope-for-the-best just doesn't seem like a maintainable approach. Easier to write and read code when you know the type of the variable.
Also better integration with tools like static analyzer.

tworec:
But this is how it works in eth lib used by us. Its a standard API.

P

pon., 13 maj 2019, 15:51 użytkownik Krigpl <notifications@github.com>
napisał:

golem_messages/utils.py
tworec: This is how it works in eth lib used by us here. Its a standard there. We support three types of input, not 'anything': string, hexstr and int.

It is usable in such a form but I can introduce dedicated params...

Krigpl: But we don't need to support three types. At this point almost everywhere we're working on strings. So just provide strings and that's it.

tworec: I needed it to support bytes representation and it seems to me I've found a pretty concise and effective way to achieve it. Much better, I think, that you can observe 5 lines below -- just expand this view and you'll see, that next util function already supports different types on the single input parameter.

This is "The Python Way":
* numpy/scipy [prominent example](https://docs.scipy.org/doc/numpy/user/basics.creation.html#converting-python-array-like-objects-to-numpy-arrays)
* pandas [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html)
* Web3 eth API [here](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.sendTransaction)

One thing you've noticed is to be corrected -- the parameter type.

Krigpl: Python way is also writing unmaintainable and illegible code, doesn't mean we should follow that.
This is the first `Union` type hint in this repository - meaning, what I said earlier, we don't do that here. We should be consistent within the same project and this case is way too minor to break current conventions.

Krigpl: I don't approve this change but I have nothing else to add and am dismissing my review since this won't be resolved.

golem_messages/factories/datastructures/tasks.py
```diff
@@ -16,10 +17,7 @@ class Meta:
         model = dt_tasks.TaskHeader
         exclude = ('requestor_public_key', )
 
-    requestor_public_key = factory.LazyFunction(
-        lambda: encode_key_id(cryptography.ECCx(None).raw_pubkey)
-    )
```
jiivan: This should still be `LazyFunction` to get different pubkeys everytime.

tests/message/tasks/test_task_to_compute.py
```diff
@@ -6,20 +6,19 @@
...
 from golem_messages import shortcuts
-
 from golem_messages.datastructures import promissory
 from golem_messages.datastructures.tasks import TaskHeader
 from golem_messages.factories.datastructures.tasks import TaskHeaderFactory
+from golem_messages.factories.helpers import random_eth_pub_key
 from golem_messages.utils import encode_hex, decode_hex
-
 from tests.message import mixins, helpers
+import factory
```
jiivan: Why move this import?

tests/message/test_concents.py
```diff
@@ -1,21 +1,19 @@
 # pylint: disable=protected-access
-from datetime import datetime, timedelta
 import math
 import time
 import unittest
+from datetime import datetime, timedelta
```
jiivan: Now it's not in alphabetical order. I preferred previous ordering.

tests/message/tasks/test_task_to_compute.py
tworec: my pylint was complaining, reverting

tests/message/test_concents.py
tworec: made by pycharm import optimizator, reverting

golem_messages/factories/datastructures/tasks.py
tworec: done

Issue #331, opened 2019-04-08
Add provider's `concent_promissory_note` to `SubtaskResultsVerify`


shadeofblue:
* [x] depends on #328 
* [x] update factories

Issue #330, opened 2019-04-08
Implement promissory note validation


shadeofblue:
add a method to validate promissory notes to `PromissoryNoteMixin`

* [x] depends on: https://github.com/golemfactory/golem-messages/issues/328

Issue #329, opened 2019-04-08
Add `promissory_note` and `concent_promissory_note` to `TaskToCompute`


shadeofblue:
Both those flags should only be present and required if `TTC.concent_enabled` is `True`

`promissory_note` is based on the provider's address and
`concent_promissory_note` is based on the Concent address

* [x] Depends on: #328 
* [x] provide methods in the TTC factory to generate messages already containing the promissory notes or modify existing ones

Issue #328, opened 2019-04-08
Implement generation of promissory note signatures


shadeofblue:
* add `PromissoryNoteMessage` mixin

PR #338, opened 2019-04-23
add promissory notes to TaskToCompute and SubtaskResultsVerify

shadeofblue:
closes: https://github.com/golemfactory/golem-messages/issues/328
closes: https://github.com/golemfactory/golem-messages/issues/329
closes: https://github.com/golemfactory/golem-messages/issues/330
closes: https://github.com/golemfactory/golem-messages/issues/331

### DIFF
--- a/golem_messages/datastructures/promissory.py
+++ b/golem_messages/datastructures/promissory.py
@@ -81,7 +81,21 @@ class PromissoryNote:
         s = (signed_message['s']).to_bytes(32, byteorder='big')
         return PromissoryNoteSig(v, r, s)
 
-    def sig_valid(self, promissory_note_sig: PromissoryNoteSig) -> bool:
+    def sig_valid(self, promissory_note_sig: typing.Optional[tuple]) -> bool:
+        if not promissory_note_sig:
+            return False
+
+        try:
+            promissory_note_sig = PromissoryNoteSig(*promissory_note_sig)
+        except TypeError as e:
+            raise TypeError(
+                "The provided promissory note sig: `{}` has wrong type. "
+                "Must be a tuple with format: {}".format(
+                    promissory_note_sig,
+                    PromissoryNoteSig._field_types,
+                )
+            ) from e
+
         address_from = eth_account.Account.recoverHash(
             message_hash=self.hash,
             vrs=promissory_note_sig
--- a/golem_messages/datastructures/promissory.py
+++ b/golem_messages/datastructures/promissory.py
@@ -104,6 +104,11 @@ class PromissorySlotMixin:
 
     CONCENT_PROMISSORY_NOTE_SIG = 'concent_promissory_note_sig'
 
+    def get_concent_promissory_note(  # noqa pylint: disable=no-self-use
+            self, deposit_contract_address: str  # noqa pylint: disable=unused-argument
+    ) -> PromissoryNote:
+        raise NotImplementedError()
+
     def sign_concent_promissory_note(
             self,
             deposit_contract_address: str,
--- a/golem_messages/datastructures/promissory.py
+++ b/golem_messages/datastructures/promissory.py
@@ -100,6 +100,13 @@ class PromissoryNote:
 
 
 class PromissorySlotMixin:
+    """
+    Defines sign and verify methods for messages containing the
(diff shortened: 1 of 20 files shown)


golem_messages/datastructures/promissory.py
```diff
@@ -0,0 +1,128 @@
...
+    'PromissoryNoteSig',
+    [
+        ('v', int),
+        ('r', bytes),
+        ('s', bytes),
+    ]
+)
+
+
+class PromissoryNote:
```
jiivan: Maybe it's a case of `datastructures.Container`?

golem_messages/datastructures/promissory.py
```diff
@@ -0,0 +1,128 @@
...
+        signed_message = eth_account.account.Account.signHash(  # noqa pylint: disable=no-value-for-parameter, line-too-long;  # pylint can't make sense of eth_utils' `@combomethod` decorator
+            message_hash=self.hash,
+            private_key=privkey
+        )
+        v = signed_message['v']
+        r = (signed_message['r']).to_bytes(32, byteorder='big')
+        s = (signed_message['s']).to_bytes(32, byteorder='big')
+        return PromissoryNoteSig(v, r, s)
+
+    def sig_valid(self, promissory_note_sig: typing.Optional[tuple]) -> bool:
```
jiivan:
```suggestion
    def verify_signature(self, promissory_note_sig: typing.Optional[tuple]) -> bool:
```
As in `Message`

golem_messages/datastructures/promissory.py
```diff
@@ -0,0 +1,128 @@
...
+        if not promissory_note_sig:
+            return False
+
+        try:
+            promissory_note_sig = PromissoryNoteSig(*promissory_note_sig)
+        except TypeError as e:
+            raise TypeError(
```
jiivan: This won't be needed if inherited from `Container`

golem_messages/utils.py
```diff
@@ -2,9 +2,21 @@
 import copy
 from uuid import UUID
 
+from eth_utils import to_checksum_address
+from ethereum.utils import sha3
+
 from golem_messages import message
 
 
+def pubkey_to_address(pubkey: str) -> str:
```
tworec: there at least two available implementations of this available:
* in bitcoin that is already a dep for GM [src](https://github.com/vbuterin/pybitcointools/blob/aeb0a2bbb8bbfe421432d776c649650eaeb882a5/bitcoin/main.py#L450)
* in pyethereum [src](https://github.com/ethereum/pyethereum/blob/b704a5c6577863edc539a1ec3d2620a443b950fb/ethereum/utils.py#L119)

maybe you can use it...

Krigpl: Bitcoin implementation returns a Bitcoin address - not what we want
this pyethereum implementation that you linked returns an address of a contract that will be created, not something that applies here
Last time I checked there was no implementation of this kind, which is weird, but that's the case.

golem_messages/datastructures/promissory.py
shadeofblue: huh? it doesn't need to be serialized/deserialized...

golem_messages/datastructures/promissory.py
shadeofblue: meh, it's a different usage than in `Message` ... it doesn't need to be similar imvho

golem_messages/datastructures/promissory.py
shadeofblue: are we serious that we want to inherit a _tuple_ from a more complicated class? ... I'd rather leave it as is

golem_messages/utils.py
```diff
@@ -2,9 +2,21 @@
 import copy
 from uuid import UUID
 
+from eth_utils import to_checksum_address
+from ethereum.utils import sha3
```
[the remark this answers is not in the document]
shadeofblue: done

golem_messages/datastructures/promissory.py
jiivan: I mistakenly thought that it's about `PromissoryNote`. Sorry.

PR #343, opened 2019-05-14
+ `PromissoryNote.subtask_id_bytes` property

shadeofblue:


golem_messages/datastructures/promissory.py
```diff
@@ -55,13 +55,17 @@ def __repr__(self):
             f"amount: {self.amount}, subtask_id: {self.subtask_id}>"
         )
 
+    @property
+    def subtask_id_bytes(self):
```
Krigpl: type hint ?

shadeofblue: --'

PR #349, opened 2019-06-05
Added envrionment prerequisites dict to TaskHeader

Wiezzel:
This field is needed for the new refactored Golem environments.


shadeofblue:
@Wiezzel is there any default value that's required for the new field? if yes, please add it to the appropriate factory so that any tests depending on it have an easier time migrating, thank you :)

Wiezzel:
@shadeofblue Nope, for now the default value is `None`.

PR #350, opened 2019-06-14
Updated all timestamps to utc timezone

maaktweluit:
Required for fixing https://github.com/golemfactory/golem/issues/4290

This change was started in https://github.com/golemfactory/golem/pull/4240 and caused the above issue

Made sure everything uses the same timezone for timestamps

### DIFF
--- a/golem_messages/factories/datastructures/tasks.py
+++ b/golem_messages/factories/datastructures/tasks.py
@@ -1,7 +1,7 @@
 # pylint: disable=too-few-public-methods
 
+import datetime
 import factory
-import time
 
 from golem_messages import cryptography
 from golem_messages.datastructures import masking
@@ -12,6 +12,10 @@ from golem_messages.factories.helpers import random_eth_pub_key
 from golem_messages.utils import encode_hex as encode_key_id
 
 
+def _date_now() -> int:
+    return int(datetime.datetime.now(tz=datetime.timezone.utc).timestamp())
+
+
 class TaskHeaderFactory(factory.Factory):
     class Meta:
         model = dt_tasks.TaskHeader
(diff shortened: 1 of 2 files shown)


golem_messages/factories/datastructures/tasks.py
```diff
@@ -1,7 +1,7 @@
 # pylint: disable=too-few-public-methods
 
 import factory
-import time
+import datetime
```
jiivan: alphabetic ordering?

maaktweluit: yes please! good catch

mfranciszkiewicz:
Could you explain what has been fixed here?

`time.time` returns seconds since epoch, which is 1.1.1970 00:00 UTC on Windows and Unix systems

maaktweluit:
@mfranciszkiewicz not sure what it exactly fixes, but it matches the changes here:
https://github.com/golemfactory/golem/pull/4240/files#diff-a2db7fb9752fab9f479467570d125244R73

i think its `time.time()` vs `time.gmtime()` but my main goal is to make the tests pass again ;)

jiivan:
> Could you explain what has been fixed here?
> 
> `time.time` returns seconds since epoch, which is 1.1.1970 00:00 UTC on Windows and Unix systems

SEE: https://docs.python.org/3/library/time.html#time.time
> Return the time in seconds since the epoch as a floating point number. The specific date of the epoch and the handling of leap seconds is **platform dependent**.

I was shocked when I learned about this, but `time.time()` does not guarantee epoch.

mfranciszkiewicz:
@jiivan Yes and it is known for Windows and Unix systems

jiivan:
I'm not a Windows expert, but on appvyor windows it returned different values than on *nixes and our local windows instances.

mfranciszkiewicz:
Since the date is predetermined geographically, why would it?

PR #353, opened 2019-08-07
Add budget to TaskToCompute

mplebanski:
A field containing subtask budget is needed for gWasm usage market.


mplebanski:
Closed. Actually it might be better to rename 'price' to 'budget'.

PR #357, opened 2019-09-04
Add task_type to TaskHeader

### LINKED ISSUE #356: Add `task_type` to TaskHeader data structure
This allows provider to determine task type on TaskHeader receival.

mplebanski:
Closes #356

### DIFF
--- a/golem_messages/datastructures/tasks.py
+++ b/golem_messages/datastructures/tasks.py
@@ -55,6 +55,7 @@ class TaskHeader(datastructures.Container):
                 fail_msg="Subtask timeout is less than 0",
             ),
         ),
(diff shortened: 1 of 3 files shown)


golem_messages/datastructures/tasks.py
```diff
@@ -55,6 +55,7 @@ class TaskHeader(datastructures.Container):
                 fail_msg="Subtask timeout is less than 0",
             ),
         ),
+        'task_type': (validators.validate_varchar128, ),
```
jiivan: Shouldn't it be an enum?

mplebanski: That would require adjusting this field every time we add a new app, which IMHO adds more work then profit

PR #359, opened 2019-09-16
Rename task_type to market_type and remove it from REQUIRED

mplebanski:
It should be merged before #358 as discussed with @mfranciszkiewicz we do not want to rely on task_type to get information about market type of the request. 

@jiivan What should I do about the release 3.12.0 which is basically a mistake? Can we take it down and make this change a 3.13.0?

### DIFF
--- a/golem_messages/constants.py
+++ b/golem_messages/constants.py
@@ -1,5 +1,6 @@
 import collections
 import datetime
+from enum import Enum
 
 from golem_messages import message
 
@@ -44,3 +45,8 @@ MSG_DELAYS = collections.defaultdict(
         message.concents.ForceReportComputedTask: (2 * MMTT + MAT),
     },
 )
+
+
+class MarketType(str, Enum):
+    BRASS_MARKET = "brass"
+    USAGE_MARKET = "usage"
--- a/golem_messages/constants.py
+++ b/golem_messages/constants.py
@@ -1,6 +1,5 @@
 import collections
 import datetime
-from enum import Enum
 
 from golem_messages import message
 
@@ -45,8 +44,3 @@ MSG_DELAYS = collections.defaultdict(
         message.concents.ForceReportComputedTask: (2 * MMTT + MAT),
     },
 )
-
-
-class MarketType(str, Enum):
-    BRASS_MARKET = "brass"
-    USAGE_MARKET = "usage"
--- a/golem_messages/datastructures/tasks.py
+++ b/golem_messages/datastructures/tasks.py
@@ -1,3 +1,4 @@
+import enum
 import functools
 import hashlib
 import logging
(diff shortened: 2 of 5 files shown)


golem_messages/constants.py
```diff
@@ -44,3 +45,8 @@
         message.concents.ForceReportComputedTask: (2 * MMTT + MAT),
     },
 )
+
+
+class MarketType(str, Enum):
+    BRASS_MARKET = "brass"
```
jiivan: Is the '_MARKET' suffix required? Please consider using `datastructures.StringEnum` and style similar to
https://github.com/golemfactory/golem-messages/blob/051492623d99135ef6928abe9791062d8105406f/golem_messages/message/tasks.py#L503-L508

golem_messages/datastructures/tasks.py
```diff
@@ -55,7 +55,7 @@ class TaskHeader(datastructures.Container):
                 fail_msg="Subtask timeout is less than 0",
             ),
         ),
-        'task_type': (validators.validate_varchar128, ),
+        'market_type': (validators.validate_varchar128, ),
```
jiivan: I think it should validate enum. You could use logic similar to this one:
https://github.com/golemfactory/golem-messages/blob/051492623d99135ef6928abe9791062d8105406f/golem_messages/message/base.py#L296-L305

golem_messages/constants.py
```diff
@@ -44,3 +45,8 @@
         message.concents.ForceReportComputedTask: (2 * MMTT + MAT),
     },
 )
+
+
+class MarketType(str, Enum):
```
jiivan: If this is only used in `TaskHeader` than I believe it should be an inner class of it.

mplebanski: Actually it would be great if one could access those values from Golem itself.

jiivan: `TaskHeader.MARKET_TYPE`?

golem_messages/datastructures/tasks.py
mplebanski: What about using deserialize_market_type and trying to convert string to an Enum. If this string is not in the enum the deserialization will fail.

golem_messages/constants.py
mplebanski: Why bind this type to a particular message class?

golem_messages/datastructures/tasks.py
jiivan: 1. `TaskHeader` is not a message. `TaskToCompute` is.
1. lno 299 does exactly that - tries to convert string/any value into enum.

golem_messages/constants.py
jiivan: > If this is only used in `TaskHeader` than I believe it should be an inner class of it.

If it's stored in some other place it might be introduced as a stand alone constant. But as far as I can see it's only used in `TaskHeader`.

mplebanski: OK

PR #363, opened 2020-01-28
Fix loading OpenSSL on macOS

zakaprov:
On MacOS, the OpenSSL version required by Golem needs to be installed
via homebrew. Golem relies on a fixed path to where the library is
installed. As part of upgrade to OpenSSL version 1.1 the installation
path has changed. On newly set up systems this prevented Golem from
finding the library. This commit appends the new installation path
to the lookup array.

### DIFF
--- a/golem_messages/cryptography.py
+++ b/golem_messages/cryptography.py
@@ -23,7 +23,11 @@ if sys.platform not in ('darwin', 'win32'):
 elif sys.platform == 'darwin':
     # FIX PATH ON OS X ()
     # https://github.com/yann2192/pyelliptic/issues/11
-    _openssl_lib_paths = ['/usr/local/Cellar/openssl/']
+    _openssl_lib_paths = [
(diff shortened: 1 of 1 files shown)


zakaprov:
https://github.com/Homebrew/homebrew-core/issues/46454

golem_messages/cryptography.py
```diff
@@ -23,7 +23,11 @@
 elif sys.platform == 'darwin':
     # FIX PATH ON OS X ()
     # https://github.com/yann2192/pyelliptic/issues/11
-    _openssl_lib_paths = ['/usr/local/Cellar/openssl/']
+    _openssl_lib_paths = [
```
maaktweluit: i think its more future proof if we do a glob for this list, something like `/usr/local/Cellar/openssl/*?`

zakaprov: I'd leave this as is for safety, especially since we need to migrate from `pyelliptic` anyway.

PR #364, opened 2020-01-29
tasks.CannotComputeTask.REASON.CannotTakeWork

jiivan:
New reason for `CannotComputeTask` message


maaktweluit:
This new reason looks a lot like `OfferCancelled`, what is the difference?

jiivan:
@maaktweluit it's meant to be used here:
https://github.com/golemfactory/golem/blob/5550f5b0d5d30fb414526280685ed8e78acfab20/golem/task/tasksession.py#L700-L702

which is triggered here:
https://github.com/golemfactory/golem/blob/5550f5b0d5d30fb414526280685ed8e78acfab20/golem/task/taskserver.py#L535-L540

